PR #12, opened 2020-10-21
Edit readme

nahuakang:
This PR makes some edits to the README.md file to improve structure as well as adding some additional info :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,9 @@ Windows users should be able to
 
 ### 1. Clone the Repository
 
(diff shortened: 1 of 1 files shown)


sunny-panchal (commented):
Thanks Nahua! The changes look good to me for the most part. The only one I'm not sure about is whether we need both the SSH and HTTPS clone instructions. We should leave the more common option and leave it to the advanced users to change that if they want. 

Https is supposed to be more widely available and accessible to all users.

guillaumebrg:
:+1:

PR #17, opened 2020-11-02
create a counting script for two exercises

antoinemrcr:
Add a script to count exercises repetitions.
Two exercises are supported : jumping jack and floor touches


guillaumebrg (commented):
Fixed conflicts and renamed a few things. Looks good to me.

Two thoughts though:
- I am not sure if everybody knows what "floor touches" are... How about squats or pushups?  
- Do we want to edit the README now? Or wait until we have the "training on temporal labels" script and mention both at the same time? 

Let's discuss tomorrow

florianletsch:
>     * I am not sure everybody knows what "floor touches" are... How about switching to squats or pushups?

I agree. I think both are fine, I'd prefer squats so that people who test it don't have to move to the floor ;)

florianletsch:
Code looks good. I'd prefer it if we can add code style checks and agree on a convention regarding 1 import per line (among other things). Should can be a different PR in the future, I think.

(Still want to test this)

guillaumebrg (approved):
Just tested it. Worked well in my setup. Thanks for switching to squats!

PR #18, opened 2020-11-02
add classifiers to the repo

antoinemrcr:
commit the classifiers head in the repo because they are small


.gitignore
```diff
@@ -3,5 +3,8 @@
 !.gitignore
 __pycache__
 *.pyc
-*.ckpt
 videos
+resources/coreml/*
+resources/strided_inflated_efficientnet.ckpt
+resources/strided_inflated_mobilenet.ckpt
```
guillaumebrg, line 9: How about we create a `resources/backbone` folder? and we ignore all the files in there?

guillaumebrg (commented):
We might need to edit in the README and explain what needs to be downloaded vs what's already there... what do you think?

@ingooooooo: we want to commit the weights of all the output layers directly in the repo. This means that we can remove them from https://20bn.com/sdk/download. Is that ok with you?

guillaumebrg (approved):
Good to go.

@ingooooooo can you merge this when you switch to `20bn-realtimenet-resources-v1.1.tgz` on the download page?

PR #19, opened 2020-11-03
[WIP] Additional changes on top of "Train classifier"

guillaumebrg:


README.md
```diff
@@ -157,8 +155,18 @@ This script expects training videos to follow this structure:
             ├── ...
         ├── ...
 ```
-- For each category you want to recognize, you should put videos in the train folder and in the valid folder.
-- All videos should have a duration of at least 3 seconds, and at least a framerate of 16 fps.
+- Two top-level folders: one for the training data, one for the validation data.
+- One sub-folder for each label with as many videos as you want (but at least one!)
+- Requirements: all videos should have a duration of at least 3 seconds, and at least a framerate of 16 fps.
```
guillaumebrg, line 160: Duplicate first frame & remove requirement?

PR #25, opened 2020-11-07
Documentation: camera.py

nahuakang:
Adding documentation and type hints to the `camera.py` classes.


guillaumebrg (approved):
Thanks :100:

Issue #26, opened 2020-11-10
Documentation: realtimenet/engine.py


nahuakang:
Now that there is documentation on `camera.py` and `video.py`, I'd like to work on documenting this file :smile: 

Will send a PR in this week.

PR #29, opened 2020-11-13
Documentation: engine.py

nahuakang:
This PR addresses [Issue #26](https://github.com/TwentyBN/20bn-realtimenet/issues/26). It adds some inline documentation and type hints to `engine.py`.

### DIFF
--- a/realtimenet/engine.py
+++ b/realtimenet/engine.py
@@ -139,18 +139,17 @@ class InferenceEngine(Thread):
 
 def run_inference_engine(
     inference_engine: InferenceEngine,
-    framegrabber: VideoStream,
+    video_stream: VideoStream,
     post_processors: List[PostProcessor],
     results_display: DisplayResults,
-    path_out: Optional[str]
-):
+    path_out: Optional[str]):
     """
     Start the video stream and the inference engine, process and display
     the prediction from the neural network.
 
     :param inference_engine:
         An instance of InferenceEngine for neural network inferencing.
-    :param framegrabber:
+    :param video_stream:
         An instance of VideoStream to feed video frames to InferenceEngine.
     :param post_processors:
         A list of subclasses of PostProcessor for post-processing neural network outputs.
@@ -170,14 +169,14 @@ def run_inference_engine(
 
     # Start threads
     inference_engine.start()
-    framegrabber.start()
+    video_stream.start()
 
     # Begin inferencing
     while True:
         frame_index += 1
 
         # Grab frame if possible
-        img_tuple = framegrabber.get_image()
+        img_tuple = video_stream.get_image()
         # If not possible, stop
         if img_tuple is None:
             break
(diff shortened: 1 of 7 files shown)


realtimenet/engine.py
```diff
@@ -67,7 +100,27 @@ def run(self):
...
+        of numpy.ndarray.
+
+        For an inference engine running 1 classifier, the output is a numpy.ndarray
+        of shape (row, col), where `col` represents the number of labels that the
+        classifier predicts.
+
+        For an inference engine running more than 1 classifier, the output is a list
+        of N numpy.ndarray, where `N` represents the number of classifiers that the
+        inference engine runs. Each of the numpy.ndarray is of shape (row, col),
+        where `col` represents the number of labels that the classifier predicts.
```
guillaumebrg, line 116: I just have a problem with using the words "classifier" / "number of labels that the classifier predicts" because this assumes a classification task. Our MET predictor, for instance, is solving a regression task and doesn't have "labels". 

I think we can rephrase this using more generic terms. How about something like this?

```
For an inference engine running a neural network producing a single output, the returned 
object is a numpy.ndarray of shape (T, C), where `C` represents the number of output 
channels while `T` represents the number of time steps and depends on the length 
of the provided input clip.

For an inference engine running a multi-output neural network, the returned object is a 
list of numpy.ndarray, one for each output.
```

realtimenet/engine.py
```diff
@@ -84,7 +137,28 @@ def infer(self, clip):
...
+    results_display: DisplayResults,
+    path_out: Optional[str]
+):
+    """
+    Start the video stream and the inference engine, process and display
+    the prediction from the neural network.
+
+    :param inference_engine:
+        An instance of InferenceEngine for neural network inferencing.
+    :param framegrabber:
```
guillaumebrg, line 153: Future PR: rename `framegrabber` to `video_stream`

guillaumebrg (commented):
Thanks @nahuakang!

I've made a few comments

guillaumebrg:
Cc #26

realtimenet/engine.py
```diff
@@ -1,14 +1,32 @@
...
 
 
 class InferenceEngine(Thread):
-
-    def __init__(self, net, use_gpu=False):
+    """
+    InferenceEngine takes in a neural network and uses it to output predictions
+    either using the local machine's CPU or GPU.
+    """
+    def __init__(self, net: Pipe, use_gpu: bool = False):
```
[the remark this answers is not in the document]
nahuakang: So I've renamed the type to `nn.Module` since all the current feature extractors and converters inherit from `nn.Module` :smile:

realtimenet/engine.py
nahuakang: Great! I edited it a bit as well :)

realtimenet/engine.py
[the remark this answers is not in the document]
guillaumebrg: Makes sense!

guillaumebrg (approved):
LGTM! :+1:

Issue #22, opened 2020-11-06
Modularize run_* scripts into a controller


nahuakang:
For the `run_*.py` scripts living under the `20bn-realtimenet/scripts/` directory, there seems to be some opportunity for modularization to extract some shared functions and be DRY.

I'll work on abstracting a `controller.py` file.

PR #30, opened 2020-11-14
Feature: Controller Class

nahuakang:
This PR draft addresses [Issue #22](https://github.com/TwentyBN/20bn-realtimenet/issues/22).

It is inspired by [@jlindermeir's hackathon app](https://github.com/jlindermeir/MediaGestureControl/blob/master/gesture_controller/controller.py) and aims to clean up the `run_*.py` scripts and offer users a slightly better experience when implementing their own scripts.

DONE:
 - ✅ Integrate `Controller` into all scripts and test the controller runs without errors.
 - ✅ Add documentation to `controller.py`.


guillaumebrg:
> Integrate Controller into all scripts and test the controller runs without errors.

I did that for `run_fitness_tracker`. I had to fight a couple of issues but it works now.

> Explore possibilities to abstract out the loading of feature extractor, classifier, and converters.
Explore possibilities to abstract out creation of post_processors.
Explore possibilities to abstract out creation of display_ops and display_results.

I don't think this is necessary. These objects differ from one script to the other... and it's okay that way. We want to keep this level of freedom

Another TODO once "Integrate Controller into all scripts and test the controller runs without errors" is done: remove `engine.run_inference_engine`

realtimenet/controller.py
```diff
@@ -0,0 +1,131 @@
...
+import numpy as np
+import torch.nn as nn
+
+
+class Controller:
+
+    def __init__(
+            self,
+            neural_network: nn.Module,
+            post_processors: Union[PostProcessor, List[PostProcessor]],
```
jlindermeir, line 20: `Controller.postprocess_prediction` is expecting `post_processors` to be a list. Either we adjust the type hint, or we add some logic to handle non-`List` inputs in the constructor.

realtimenet/controller.py
```diff
@@ -0,0 +1,131 @@
...
+            results_display: DisplayResults,
+            camera_id: int,
```
jlindermeir, line 22: I think we can set a default of `0` here, since most people only have one camera available.

realtimenet/controller.py
```diff
@@ -0,0 +1,131 @@
...
+            path_in: str = None,
+            path_out: str = None,
+            use_gpu: bool = True):
+
+        video_source = VideoSource(
+            camera_id=camera_id,
+            size=self.inference_engine.expected_frame_size,
+            filename=path_in
+        )
+        self.video_stream = VideoStream(video_source, self.inference_engine.fps)
```
jlindermeir, line 32: `self.inference_engine` is used before assignment here.

realtimenet/controller.py
```diff
@@ -0,0 +1,131 @@
...
+
+                # Get predictions
+                prediction = self.inference_engine.get_nowait()
+
+                prediction_postprocessed = self.postprocess_prediction(prediction)
+
+                self.display_prediction(img, prediction_postprocessed)
+
+            except Exception as runtime_error:
+                break
```
jlindermeir, line 79: The entire point of writing this controller war to continuously obtain predictions in a loop, which could then be used for some task. I think we should add to option to specify a callback in the constructor, which is called on every loop with the new predictions as an argument. Without further modification, this class is only useful for displaying results.

I am also not entirely happy with the structure of this block, there is a lot going on. Maybe we can refactor `frame_index` and `clip` into attributes which are initialized during `_start_inference`. We could then encapsulate most of the loop into something like `_get_predictions`. I think that would greatly increase readability.

realtimenet/controller.py
```diff
@@ -0,0 +1,131 @@
...
+
+            # Press escape to exit
+            if cv2.waitKey(1) == 27:
+                break
```
jlindermeir, line 83: I think this only works if `DisplayResults` is provieded.

jlindermeir (commented):
@nahuakang Thanks for taking the initiative on this, I think providing an interface for obtaining predictions is a good idea and would increase usability of this repo.
I've left a few comments below.

realtimenet/controller.py
```diff
@@ -0,0 +1,174 @@
...
+            results_display: DisplayResults,
+            callbacks: Optional[List[Callable]] = None,
+            camera_id: int = 0,
+            path_in: str = Optional[None],
+            path_out: str = Optional[None],
+            use_gpu: bool = True):
+        """
+        :param neural_network:
+            The neural network that produces the predictions for the camera image.
+            TODO: Be more specific here
```
guillaumebrg, line 31: Remaining TODO. On a related note, `neural_network: nn.Module` might be a bit misleading because not every `nn.Module` will work as `InferenceEngine` expects `net` to define certain attrivutes/methods (`net.fps`, `net.step_size`, `net.preprocess` etc...).  We should create an abstract RealtimeNeuralNet class that clarifies the requirements. Should be done in another PR; let's leave the TODO here for now.

guillaumebrg (approved):
Great work!

PR #31, opened 2020-11-16
Feature: pycodestyle

nahuakang:
Quickly adding `pycodestyle` to the repo (config copied from our very own FE).

This way we can call `pycodestyle` in the project root directory and check if our code is complying to the code style guideline. Helps us write clearer code :smile:

### DIFF
--- a/realtimenet/setup.cfg
+++ b/realtimenet/setup.cfg
@@ -1,0 +1,6 @@
+# Config for pycodestyle
+[pycodestyle]
+max-line-length = 120
+exclude = resources/*
+count = True
+statistics = True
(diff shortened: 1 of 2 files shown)


florianletsch:
Does the current repo pass the checks? My guess is not.
I think we should only merge this when it actually means a "Green" check and make it mandatory to pass for future PRs.

nahuakang:
No the current repo has quite a handful of failures :smile: Makes sense. Let's wait until the repo is fixed (I might try this weekend).

guillaumebrg:
I think we can close this now. `pycodestyle` was added with #81

PR #33, opened 2020-11-20
Fix: coreml conversion script

YasheshSavani:
This PR will change the `requirement.txt` in which the `h5py==2.10.0` is added. It will resolve the error of coreml conversion because KerasAPI does not support h5py>3.0.0.


sunny-panchal (approved):
Created a fresh environment with these requirements and was able to convert the model :+1:

Issue #27, opened 2020-11-11
Better reporting of the inference engine fps


guillaumebrg:
Our scripts don't do a good job at warning the user when the script isn't running in optimal conditions. We do have a few prints (e.g. *** Unused frames ***) but these are easy to overlook, especially if you're focused on the debug window. Should the debug window report the inference engine fps? What about the camera fps?

Related issue: #16

PR #36, opened 2020-11-24
[Fix] FPS counting for camera and inference engine

YasheshSavani:
This PR is related with the issue #27.
It will add inference engine fps and camera fps in debug window for better reporting which will show users that when the script isn't running in optimal conditions.


realtimenet/display.py
```diff
@@ -171,6 +171,10 @@ def show(self, img: np.ndarray, display_data: dict) -> np.ndarray:
         # Add black borders
         img = cv2.copyMakeBorder(img, self.border_size, 0, 0, 0, cv2.BORDER_CONSTANT)
 
+        # Show FPS on the video screen
+        put_text(img, "Camera FPS: {}".format(camera_fps), (10, 18))
+        put_text(img, "Inference Engine FPS: {}".format(inference_engine_fps), (360, 18))
```
YasheshSavani, line 176: I want help with this what should we call it. Any suggestions?

realtimenet/engine.py
```diff
@@ -194,16 +198,26 @@ def run_inference_engine(
 
         # Get predictions
         prediction = inference_engine.get_nowait()
+        model_inference_end_time = time.time()
+
+        # Inference engine frame rate
+        inference_engine_fps = (1/(model_inference_end_time - model_inference_start_time))/inference_engine.step_size
+        model_inference_start_time = model_inference_end_time
```
guillaumebrg, line 205: Getting to this point does not mean that a new prediction was produced because `inference_engine.get_nowait()` returns `None` in case the prediction wasn't ready. Here is what I think should be done:
```
if predictions is not None:
        model_inference_end_time = time.time()
        inference_engine_fps = 1 / (model_inference_end_time - model_inference_start_time)
        model_inference_start_time = model_inference_end_time
```
(no need to divide by `inference_engine.step_size` then)

guillaumebrg (commented):
Made a few comments. In addition to those, I would suggest to display a running average of the fps rather than the instantaneous value (it's changing too quickly otherwise, I find it distracting)

realtimenet/display.py
```diff
@@ -171,6 +171,10 @@ def show(self, img: np.ndarray, display_data: dict) -> np.ndarray:
         # Add black borders
         img = cv2.copyMakeBorder(img, self.border_size, 0, 0, 0, cv2.BORDER_CONSTANT)
 
+        # Show FPS on the video screen
+        put_text(img, "Camera FPS: {:.2f}".format(camera_fps), (10, 18))
+        put_text(img, "Model FPS: {:.1f}".format(inference_engine_fps), (360, 18))
```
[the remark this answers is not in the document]
YasheshSavani: It can be done

realtimenet/engine.py
```diff
@@ -165,11 +166,14 @@ def run_inference_engine(
     display_error = None
     video_recorder = None
     video_recorder_raw = None
+    previous_camera_frame_time = 0
 
     # Start threads
     inference_engine.start()
     video_stream.start()
 
+    model_inference_start_time = time.time()
```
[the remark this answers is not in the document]
YasheshSavani: will checkout  that and do the necessary change.

realtimenet/display.py
```diff
@@ -153,7 +153,7 @@ def __init__(self, title: str, display_ops: List[BaseDisplay], border_size: int
         self.display_ops = display_ops
         self.border_size = border_size
 
-    def show(self, img: np.ndarray, display_data: dict) -> np.ndarray:
+    def show(self, img: np.ndarray, display_data: dict, camera_fps: float, inference_engine_fps: float) -> np.ndarray:
```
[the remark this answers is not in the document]
YasheshSavani: Good idea

realtimenet/engine.py
YasheshSavani: I tried and tested it, sometimes it goes above 4 ranging from 5 or 16. What should we do for that?

guillaumebrg:
Hey @YasheshSavani! #30 was merged... so this will have to be re-worked a bit. Anyway, I've been thinking a bit more about this and I think we should try to move most of the logic inside a new `ResultsDisplay` class. Let's discuss on Monday

realtimenet/display.py
```diff
@@ -133,6 +134,47 @@ def display(self, img, display_data):
         return img
 
 
+class DisplayFPS(BaseDisplay):
+    """
+    Display camera fps and inference engine fps on debug window.
+    """
+
+    def __init__(self, neural_network, y_offset=20):
```
guillaumebrg, line 142: Instead of passing the whole neural net, how about we add two arguments called `expected_camera_fps` and `expected_inference_fps`? I think that's everything we need. These arguments should be optional and default to `None`. If left to `None`, `self.running_delta_time_inference` and `self.running_delta_time_camera`  can be initialized to 0 and the text color will stay green.

realtimenet/display.py
```diff
@@ -133,6 +134,47 @@ def display(self, img, display_data):
...
+        super().__init__(y_offset)
+        self.net = neural_network
+
+        self.fps_update_rate = 0.1
```
guillaumebrg, line 146: I think we can call it `self.update_rate` now that we have a specific class for it

realtimenet/display.py
```diff
@@ -133,6 +134,47 @@ def display(self, img, display_data):
...
+        inference_engine_fps = 1. / self.running_delta_time_inference
+
+        # Camera FPS counting
+        delta = (now - self.last_update_time_camera)
+        self.running_delta_time_camera += self.fps_update_rate * (delta - self.running_delta_time_camera)
+        camera_fps = 1. / self.running_delta_time_camera
+        self.last_update_time_camera = now
+
+        # Text color change if inference engine fps go below certain range
+        text_color = (0, 255, 0) if inference_engine_fps > self.net.step_size * 0.75 else (0, 0, 255)
```
guillaumebrg, line 169: Let's add a class attribute for this `0.75`. How about `self.low_performance_rate`?

guillaumebrg (commented):
Thanks for addressing my comment! :)

Made a bunch of minor comments... Almost there :crossed_fingers:

guillaumebrg (approved):
Made some small display changes

PR #46, opened 2020-11-29
Abstract Neural Network Class to replace nn.Module as type hint

nahuakang:
Per Guillaume's comment:
The parameter inside `Controller` class: `neural_network: nn.Module` might have a misleading type hint because not every `nn.Module` will work as a parameter for `InferenceEngine` expects those with certain attributes/methods (`net.fps`, `net.step_size`, `net.preprocess` etc...). We should create an abstract `RealtimeNeuralNet` class that clarifies the requirements.

### DIFF
--- a/realtimenet/controller.py
+++ b/realtimenet/controller.py
@@ -7,7 +7,7 @@ from realtimenet.camera import VideoSource
 from realtimenet.camera import VideoStream
 from realtimenet.display import DisplayResults
 from realtimenet.engine import InferenceEngine
-from realtimenet.downstream_tasks.nn_utils import RealtimeNN
+from realtimenet.downstream_tasks.nn_utils import RealtimeNeuralNetwork
 from realtimenet.downstream_tasks.postprocess import PostProcessor
 
 import cv2
@@ -17,7 +17,7 @@ import numpy as np
 class Controller:
     def __init__(
             self,
-            neural_network: RealtimeNN,
+            neural_network: RealtimeNeuralNetwork,
             post_processors: Union[PostProcessor, List[PostProcessor]],
             results_display: DisplayResults,
             callbacks: Optional[List[Callable]] = None,
--- a/realtimenet/controller.py
+++ b/realtimenet/controller.py
@@ -7,17 +7,17 @@ from realtimenet.camera import VideoSource
 from realtimenet.camera import VideoStream
 from realtimenet.display import DisplayResults
 from realtimenet.engine import InferenceEngine
+from realtimenet.downstream_tasks.nn_utils import RealtimeNeuralNet
 from realtimenet.downstream_tasks.postprocess import PostProcessor
 
 import cv2
 import numpy as np
-import torch.nn as nn
 
 
 class Controller:
     def __init__(
             self,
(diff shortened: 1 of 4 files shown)


guillaumebrg (commented):
Just skimmed it and it looks good to me. I'll try to think whether feature extractors should also inherit from it

realtimenet/downstream_tasks/nn_utils.py
```diff
@@ -1,7 +1,48 @@
+import numpy as np
 import torch.nn as nn
 
+from typing import Tuple
 
-class Pipe(nn.Module):
+
+class RealtimeNeuralNet(nn.Module):
```
sunny-panchal, line 7: A name to consider: RealtimeNN

realtimenet/downstream_tasks/nn_utils.py
```diff
@@ -1,7 +1,48 @@
...
+    """
+    RealtimeNeuralNet is the abstract class for all neural networks which
+    are to be fed into an instance of the InferenceEngine.
```
sunny-panchal, line 10:
```suggestion
    RealtimeNeuralNet is an abstract class for neural networks used in InferenceEngine.
```

sunny-panchal (commented):
some minor comments:

guillaumebrg:
I think I preferred `RealtimeNeuralNet` to `RealtimeNN` :see_no_evil: Do you both prefer `RealtimeNN`?

I think feature extractors, e.g. `StridedInflatedMobileNetV2`, should inherit from it too.

nahuakang:
I have very little preference at this stage since we might change the name of the repo. But on second thought, explicit naming is better than implicit naming :smile: 

@guillaumebrg I currently feel like `feature_converters` have quite a few different inheritance from `torch`? Without further knowledge, it seems a bit harder to streamline these converters :thinking:

realtimenet/downstream_tasks/nn_utils.py
```diff
@@ -1,7 +1,47 @@
+import numpy as np
 import torch.nn as nn
 
+from typing import Tuple
 
-class Pipe(nn.Module):
+
+class RealtimeNeuralNetwork(nn.Module):
+    """
+    RealtimeNeuralNet is the abstract class for all neural networks used in InferenceEngine.
```
guillaumebrg, line 9: Class is called `RealtimeNeuralNetwork` now, docstring says `RealtimeNeuralNet`

realtimenet/controller.py
```diff
@@ -7,17 +7,17 @@
...
 import cv2
 import numpy as np
-import torch.nn as nn
 
 
 class Controller:
     def __init__(
             self,
-            neural_network: nn.Module,
+            neural_network: RealtimeNeuralNetwork,
```
guillaumebrg, line 20: Should we edit the docstring and remove the TODO now?

guillaumebrg (commented):
Let's not worry about `feature_converters`, these can inherit from `nn.Module`.

Made 2 small comments, looks good otherwise.

guillaumebrg (approved):
:1st_place_medal:

PR #47, opened 2020-12-01
[bugfix] Remove softmax layer during finetuning

guillaumebrg:


antoinemrcr (commented):
I can be ok with that, but would not it be better if we had a parameter for the logisticResgression to include or not the softmax at initialization ?

antoinemrcr (approved):
LGTM

PR #48, opened 2020-12-04
Format folder structures in `tree` style

corneliusboehm:
For the folder structures in README, use the formatting that is also output by the `tree` command.


guillaumebrg (approved):
Thanks!

PR #51, opened 2020-12-10
VideoCapture script - recorded FPS

manikd31:
This PR updates the FPS calculated as is in case it's more than 16, or 16 otherwise.
Previously got merged without the updated value.


antoinemrcr (approved):
LGTM

PR #56, opened 2020-12-14
[Feature] AV-3999 : Do not end on index error

YasheshSavani:
Ticket: https://20bn.atlassian.net/browse/AV-3999

Description:
This PR will add an update to `annotation.py` script to fix the issue of getting index error when all videos are annotated in a folder. Now, as an update after the last video's annotations submission it will redirect to the list of videos page.

### DIFF
--- a/scripts/tools/annotation/annotation.py
+++ b/scripts/tools/annotation/annotation.py
@@ -13,7 +13,7 @@ from joblib import dump, load
 
 dataset_path = '/home/amercier/code/20bn-realtimenet/fitness_tl_benchmark/'
 split = 'train'
-label = 'with_racket'
+label = 'Spider Man Pushup'
 folder = os.path.join(dataset_path, f'videos_{split}', label)
 
 # out_folder = '/home/amercier/code/20bn-realtimenet/annotation/0/'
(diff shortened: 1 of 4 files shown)


scripts/tools/annotation/annotation.py
```diff
@@ -11,9 +11,9 @@
 from sklearn.linear_model import LogisticRegression
 from joblib import dump, load
 
-dataset_path = '/home/amercier/code/20bn-realtimenet/fitness_tl_benchmark/'
+dataset_path = '/home/twentybn/Code/sense/dataset/'
```
antoinemrcr, line 14: should not commit this change nor the next one

scripts/tools/annotation/prepare_annotation.py
```diff
@@ -19,9 +19,9 @@
 # Create inference engine
 inference_engine = engine.InferenceEngine(feature_extractor, use_gpu=True)
 
-dataset_path = '/home/amercier/code/20bn-realtimenet/fitness_tl_benchmark/'
+dataset_path = '/home/twentybn/Code/sense/dataset/'
```
antoinemrcr, line 22: do not commit this please

scripts/tools/annotation/annotation.py
YasheshSavani: ok

scripts/tools/annotation/prepare_annotation.py
YasheshSavani: ok

PR #57, opened 2020-12-14
[Fix] AV-3998: Check annotation before training

YasheshSavani:
**Ticket**: https://20bn.atlassian.net/browse/AV-3998

**Description:**

This PR will fix the problem of `shape not matching` or the error `This LogisticRegression instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator` if you click on button `Retrain model`. It will check the annotation if they exist before training on it. For logistic regression, labels will be set iff the relevant classes exists in annotations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@ __pycache__
 videos
 resources/coreml/*
 resources/backbone/*
-
+dataset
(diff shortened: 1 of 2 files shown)


antoinemrcr (approved):
LGTM

PR #59, opened 2020-12-15
Release lite models

guillaumebrg:
This PR is preparing the release of our lite models. It introduces a few concepts that should make it easier for the user to switch between different models when running a demo. Demos that can be run by multiple models now includesa section like this at the top of the script:
```
SUPPORTED_MODEL_CONFIGURATIONS = [
    ModelConfig('StridedInflatedMobileNetV2', 'pro', ['met_converter']),
    ModelConfig('StridedInflatedEfficientNet', 'pro', ['met_converter']),
    ModelConfig('StridedInflatedMobileNetV2', 'lite', ['met_converter']),
    ModelConfig('StridedInflatedEfficientNet', 'lite', ['met_converter']),
]
```
indicating which models are supported by this script. It is possible to select which model to use from the command line (see `--model_name` and `--model_version` arguments). When these arguments are not used (default behavior), it will loop over `SUPPORTED_MODEL_CONFIGURATIONS`, from top to bottom, until it can find a model configuration for which the user has all necessary files. With the example above, this means that the script will try to load "pro" weights first and will only fallback to "lite" weights if "pro" weights were not found.

Please, also check out `models.yaml`. This is my suggestion to map classifier weights to their corresponding backbone weights. This is what we will have to edit when we add support to another demo.

I've left a bunch of TODOs here and there. We should also edit the README I guess.


scripts/run_gesture_recognition.py
```diff
@@ -7,46 +7,66 @@
...
                              [--use_gpu]
   run_gesture_recognition.py (-h | --help)
 
 Options:
   --path_in=FILENAME         Video file to stream from
   --path_out=FILENAME        Video file to stream to
   --title=TITLE              This adds a title to the window display
+  --model_name=NAME          Name of the model to be used.
+  --model_version=VERSION    Version of the model to be used.
+  --use_gpu                  Whether to run inference on the GPU or not.
```
antoinemrcr, line 21: should we default to one model version ?

guillaumebrg: It's not necessary:
> When these arguments are not used (default behavior), it will loop over SUPPORTED_MODEL_CONFIGURATIONS, from top to bottom, until it can find a model configuration for which the user has all necessary files

Adding a default model_name/model_version wouldn't add anything

sense/loading.py
```diff
@@ -0,0 +1,125 @@
+import json
+import os
+import torch
+import yaml
+
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+with open(os.path.join(os.path.dirname(__file__), 'models.yml')) as f:  # TODO: check if this is the best way of doing this
```
sunny-panchal, line 10: @corneliusboehm any thoughts on this?

sunny-panchal (commented):
Finished making the remaining changes and it works as intended :+1:

sense/loading.py
corneliusboehm: Apparently you can get more complex (https://stackoverflow.com/a/4060259/6611214), but this looks ok to me

guillaumebrg:
Will fix conflicts tomorrow

examples/run_fitness_rep_counter.py
```diff
@@ -7,25 +7,36 @@
...
                              [--use_gpu]
   run_fitness_rep_counter.py (-h | --help)
 
 Options:
   --path_in=FILENAME              Video file to stream from
   --path_out=FILENAME             Video file to stream to
   --title=TITLE                   This adds a title to the window display
+  --model_name=NAME          Name of the model to be used.
+  --model_version=VERSION    Version of the model to be used.
+  --use_gpu                  Whether to run inference on the GPU or not.
```
guillaumebrg, line 21: Misalignment

examples/run_fitness_rep_counter.py
```diff
@@ -35,22 +46,28 @@
...
-    feature_extractor.load_weights_from_resources('backbone/strided_inflated_efficientnet.ckpt')
-    feature_extractor.eval()
+    # Load weights
+    selected_config, weights = get_relevant_weights(
+        SUPPORTED_MODEL_CONFIGURATIONS,
+        model_name,
+        model_version
+    )
+
+    # Load backbone extractor
```
guillaumebrg, line 60: "backbone extractor"

examples/run_fitness_tracker.py
```diff
@@ -45,29 +60,34 @@
...
-    feature_extractor.load_weights_from_resources('backbone/strided_inflated_mobilenet.ckpt')
-    feature_extractor.eval()
+    # Load weights
+    selected_config, weights = get_relevant_weights(
+        SUPPORTED_MODEL_CONFIGURATIONS,
+        model_name,
+        model_version
+    )
+
+    # Load backbone extractor
```
guillaumebrg, line 74: "backbone extractor"

examples/run_gesture_recognition.py
```diff
@@ -7,50 +7,69 @@
...
-    feature_extractor.load_weights_from_resources('backbone/strided_inflated_efficientnet.ckpt')
-    feature_extractor.eval()
+    # Load weights
+    selected_config, weights = get_relevant_weights(
+        SUPPORTED_MODEL_CONFIGURATIONS,
+        model_name,
+        model_version
+    )
+
+    # Load backbone extractor
```
guillaumebrg, line 62: backbone extractor

sense/loading.py
```diff
@@ -0,0 +1,175 @@
...
+                                'To download, please go to '
+                                'https://20bn.com/licensing/sdk/evaluation and follow the '
+                                'instructions.'.format(checkpoint_path))
+
+
+def load_backbone_weights(checkpoint_path: str):
+    """
+    Load backbone weights from a checkpoint file, unless Travis is used. Raises an error pointing
+    to the SDK page in case weights
+    are missing.
```
guillaumebrg, line 142: Misalignment

sense/loading.py
```diff
@@ -0,0 +1,175 @@
+import json
```
guillaumebrg, line 1: TODO: add unittests to cover classes/functions from this file

guillaumebrg (commented):
A bunch of notes to myself... will address these soon

sense/loading.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def load_weights_from_resources(checkpoint_path: str):
+    """
+    Load weights from a checkpoint file located in the resources folder.
+
+    :param checkpoint_path:
+        A string representing the absolute/relative path to the checkpoint file.
+    """
+    checkpoint_path = os.path.join(RESOURCES_DIR, checkpoint_path.split(f'resources{os.sep}')[-1])
```
corneliusboehm, line 127: One more reason to remove `resources/` from all paths in the config file

sense/loading.py
```diff
@@ -0,0 +1,174 @@
...
+    Load backbone weights from a checkpoint file, unless Travis is used. Raises an error pointing
+    to the SDK page in case weights are missing.
+
+    :param checkpoint_path:
+        A string representing the absolute/relative path to the checkpoint file.
+    """
+    if not using_travis():
+        return load_weights_from_resources(checkpoint_path)
+    else:
+        print('Weights are not loaded on Travis.')
```
corneliusboehm, line 149: Is it really necessary to have this method wrap `load_weights_from_resources`?

sense/loading.py
```diff
@@ -0,0 +1,174 @@
...
+        backbone_network.load_state_dict(weights)
+    backbone_network.eval()
+    return backbone_network
+
+
+def using_travis():
+    """
+    Returns True if Travis is currently being used.
+    """
+    return os.getenv('TRAVIS', False) == 'true'
```
corneliusboehm, line 174: How about making this a constant `TRAVIS_ENV` on top of the file? Or does it change at runtime?

corneliusboehm (commented):
Looking good :slightly_smiling_face: Some general feedback that we could address now or later:

I would be in favor of empowering the `ModelConfig` class a little more, e.g. by moving some parts from `get_relevant_weights` or even the model building into it.

Also it would be nice if we could simplify the model setup API even more, because currently this much code is still required to setup a model with a default config and we have it as kind of boilerplate in each example script:
```python
    # Load weights
    selected_config, weights = get_relevant_weights(
        SUPPORTED_MODEL_CONFIGURATIONS,
        model_name,
        model_version
    )

    # Load backbone network
    backbone_network = build_backbone_network(selected_config, weights['backbone'])

    # Create a logistic regression classifier
    gesture_classifier = LogisticRegression(num_in=backbone_network.feature_dim,
                                            num_out=30)
    gesture_classifier.load_state_dict(weights['gesture_recognition'])
    gesture_classifier.eval()

    # Concatenate backbone network and logistic regression
    net = Pipe(backbone_network, gesture_classifier)
```
I'm imagining shortening it to this, which would only require us to store the class and dimensions of the classifier in the config as well:
```python
    # Setup model
    net = load_model(
        SUPPORTED_MODEL_CONFIGURATIONS,
        model_name,
        model_version
    )
```

sense/models.yml
```diff
@@ -0,0 +1,29 @@
+StridedInflatedEfficientNet:
+
+  lite:
+    backbone: resources/backbone/strided_inflated_efficientnet_lite.ckpt
+    fitness_activity_recognition: resources/fitness_activity_recognition/strided_inflated_efficientnet_lite_logistic_regression.ckpt
+    gesture_recognition: resources/gesture_recognition/strided_inflated_efficientnet_lite_logistic_regression.ckpt
+    met_converter: resources/calorie_estimation/strided_inflated_efficientnet_lite_met_converter.ckpt
```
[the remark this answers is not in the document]
guillaumebrg: Suggested a few changes here: 5ea2f67

sense/loading.py
```diff
@@ -0,0 +1,174 @@
...
+                load_fn = load_backbone_weights if name == 'backbone' else load_weights
+                weights[name] = load_fn(path)
+
+            return model_config, weights
+        else:
+            print(f'Could not find at least one of the following files:\n{path_weights_string}')
+
+    raise Exception('ERROR - Weights file missing. To download, please go to '
+                    'https://20bn.com/licensing/sdk/evaluation and follow the '
+                    'instructions.')
```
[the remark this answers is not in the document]
guillaumebrg: I think this should be fixed now - `load_weights` was used somewhere, instead of `load_weights_from_resources`

sense/loading.py
guillaumebrg: It is useful at the time where I write this. Integration tests would complain on Travis without this. But maybe this can go when we refactor `get_relevant_weights`

guillaumebrg:
@corneliusboehm Thanks for the review! I've addressed the issues that were easy to fix.

Moving `get_relevant_weights`  into `ModelConfig` would probably make sense and I think there are a few things that could be simplified with regards to how we handle loading weights on Travis (not sure we need this many `running_on_travis` conditions)...

I haven't made up my mind yet about your `load_model` suggestion. One disadvantage I see is that it would hide high-level information about the architecture (e.g. "the model for this example is simply a backbone network + a fully connected layer with X classes") and, as a user, I kind of liked that this was exposed in the example scripts. Not sure... Maybe we can discuss it with a larger round of people and see what they prefer?

sense/loading.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def prepend_resources_path(checkpoint_path):
+    """
+    Prepend the absolute resources path to the provided path.
+    """
+    return os.path.join(RESOURCES_DIR, checkpoint_path.split(f'resources{os.sep}')[-1])
```
corneliusboehm, line 116: The split on `resources/` should not necessary, now that  the paths no longer contain it.

corneliusboehm (approved):
Thanks for the changes. I confirmed that everything is working :+1: 

We can talk about the other ideas in a larger round.

tools/conversion/convert_to_tflite.py
```diff
@@ -61,17 +64,17 @@
...
-        "corresponding_backbone": "efficientnet",
+        "weights_file": MODELS[EFFICIENTNET]['pro']['gesture_recognition'],
+        "corresponding_backbone": EFFICIENTNET,
     },
     "efficient_net_fitness_activity_recognition": {
         "config_file": "tools/conversion/cfg/logistic_regression.cfg",
         "placeholder_values": {"NUM_CLASSES": "81"},
-        "weights_file": "resources/fitness_activity_recognition/efficientnet_logistic_regression.ckpt",
-        "corresponding_backbone": "efficientnet",
+        "weights_file": MODELS[EFFICIENTNET]['pro'],
```
corneliusboehm, line 76:
```suggestion
        "weights_file": MODELS[EFFICIENTNET]['pro']['fitness_activity_recognition'],
```

guillaumebrg: Thanks!

corneliusboehm (dismissed):

PR #63, opened 2020-12-16
[Feature] AV-3984: Create script for video flipping horizontally

YasheshSavani:
Ticket: https://20bn.atlassian.net/browse/AV-3984

Description:
This PR will add a script named `flip_video.py` inside `scripts/tools` folder. This script will help user to flip their videos horizontally to create more data.

### DIFF
--- a/scripts/tools/flip_video.py
+++ b/scripts/tools/flip_video.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+"""
+This script helps to flip videos horizontally to create videos of opposite side.
+
+Usage:
+  flip_video.py --path_in=PATH_IN --path_out=PATH_OUT
+  flip_video.py (-h | --help)
+
+Options:
+  --path_in=PATH_IN     Full path to the videos folder
+  --path_out=PATH_OUT   Full path to save flipped videos
+"""
+
+import ffmpeg
+from docopt import docopt
+from pathlib import Path
+import os
+
+if __name__ == '__main__':
+    args = docopt(__doc__)
+    videos_path_in = Path(args['--path_in'])
+    videos_path_out = Path(args['--path_out'])
+
+    video_suffix = '.mp4'
+
+    os.makedirs(videos_path_out, exist_ok=True)
+    for directory in videos_path_in.iterdir():
+
+        print(f'Processing videos in folder: {directory.name}')
+        output_folder = videos_path_out / directory.name
+        os.makedirs(output_folder, exist_ok=True)
+
+        for video in directory.iterdir():
+            print(f'Processing video: {video.name}')
+
+            original_video_name = video.stem
+            flipped_video_name = original_video_name + '_flipped' + video_suffix
+
+            # Original video as input
+            original_video = ffmpeg.input(directory / video.name)
+            # Do horizontal flip
+            flipped_video = ffmpeg.hflip(original_video)
+            # Get flipped video output
+            flipped_video_output = ffmpeg.output(flipped_video, filename=output_folder / flipped_video_name)
+            # Run to render and save video
+            ffmpeg.run(flipped_video_output)
+
+    print("Processing done!")
--- a/scripts/tools/flip_video.py
+++ b/scripts/tools/flip_video.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 """
-This script helps to flip videos horizontally to create videos of opposite side.
+This script helps to flip videos horizontally for data augmentation. 
+
+Generally, it can be used to quickly double the size of your dataset, or, in the case where you've collected data for an action performed on a specific side, you can flip these videos and use them to classify the opposite side.
 
 Usage:
   flip_video.py --path_in=PATH_IN --path_out=PATH_OUT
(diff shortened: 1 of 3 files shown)


scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+
+        print(f'Processing videos in folder: {directory.name}')
+        output_folder = videos_path_out / directory.name
+        os.makedirs(output_folder, exist_ok=True)
+
+        for video in directory.iterdir():
+            print(f'Processing video: {video.name}')
+
+            original_video_name = video.stem
+            flipped_video_name = original_video_name + '_flipped' + video_suffix
```
YasheshSavani, line 37: Need suggestion on naming of output video.

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+"""
+This script helps to flip videos horizontally to create videos of opposite side.
```
sunny-panchal, line 3:
```suggestion
This script helps to flip videos horizontally for data augmentation. 

Generally, it can be used to quickly double the size of your dataset, or, in the case where you've collected data for an action performed on a specific side, you can flip these videos and use them to classify the opposite side.
```

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+
+Options:
+  --path_in=PATH_IN     Full path to the videos folder
+  --path_out=PATH_OUT   Full path to save flipped videos
+"""
+
+import ffmpeg
+from docopt import docopt
+from pathlib import Path
+import os
```
sunny-panchal, line 17:
```suggestion
import ffmpeg
import os

from docopt import docopt
from pathlib import Path
```

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+
+if __name__ == '__main__':
+    args = docopt(__doc__)
+    videos_path_in = Path(args['--path_in'])
+    videos_path_out = Path(args['--path_out'])
+
+    video_suffix = '.mp4'
```
sunny-panchal, line 24: Do we need to hard-code this? Can we just detect and preserve the same extension as the original?

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+
+import ffmpeg
+from docopt import docopt
+from pathlib import Path
+import os
+
+if __name__ == '__main__':
+    args = docopt(__doc__)
+    videos_path_in = Path(args['--path_in'])
+    videos_path_out = Path(args['--path_out'])
```
sunny-panchal, line 22: Should we set a default path out as the path in folder?

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+
+    os.makedirs(videos_path_out, exist_ok=True)
+    for directory in videos_path_in.iterdir():
+
+        print(f'Processing videos in folder: {directory.name}')
+        output_folder = videos_path_out / directory.name
+        os.makedirs(output_folder, exist_ok=True)
+
+        for video in directory.iterdir():
+            print(f'Processing video: {video.name}')
```
sunny-panchal, line 34: Inconsistent empty lines

scripts/tools/flip_video.py
sunny-panchal: This works. Could combine with the line above instead of setting a new variable for `original_video_name`

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+            # Original video as input
+            original_video = ffmpeg.input(directory / video.name)
+            # Do horizontal flip
+            flipped_video = ffmpeg.hflip(original_video)
+            # Get flipped video output
+            flipped_video_output = ffmpeg.output(flipped_video, filename=output_folder / flipped_video_name)
+            # Run to render and save video
+            ffmpeg.run(flipped_video_output)
+
+    print("Processing done!")
```
sunny-panchal, line 48:
```suggestion
    print("Processing complete!")
```

scripts/tools/flip_video.py
```diff
@@ -0,0 +1,48 @@
...
+import os
+
+if __name__ == '__main__':
+    args = docopt(__doc__)
+    videos_path_in = Path(args['--path_in'])
+    videos_path_out = Path(args['--path_out'])
+
+    video_suffix = '.mp4'
+
+    os.makedirs(videos_path_out, exist_ok=True)
```
sunny-panchal, line 26: pathlib also has a method to create a directory if you want to use the same one

sunny-panchal (commented):
It didn't work when I tried it. I should be able to pass in the path to a folder containing videos I want to flip and get them saved in another folder. 

If I pass in a path to a specific folder that I want to flip the videos in, i.e. `../videos_train/background` it breaks because the loop on line `27` is expecting sub-folders. 

If I pass in the path of `videos_train`, the script performs the flip but crashes when it comes time to save the video. For `--path_out` I provided `../videos_train_flipped/`

scripts/tools/flip_video.py
YasheshSavani: I think I tried once with `.avi` format but then it did not work in case of training. I don't remember perfectly though.

YasheshSavani:
New PR created #66 with rebasing to master.

PR #64, opened 2020-12-17
Add digit recognition example

corneliusboehm:


florianletsch:
I'm pretty sure this is unrelated to your example, but in a fresh installation of the repo and environment, I can't get the demo to work. It prints two lines, does not open a window, does not show an error, and simply terminates.

Have you seen this before?

```
$ PYTHONPATH=./ python scripts/run_digit_recognition.py --camera_id=1
Starting inference
Stopping inference

$ 

```

corneliusboehm:
Don't think I have seen this :confused: Do any of the other demos work?

sense/display.py
```diff
@@ -10,8 +10,14 @@
 FONT = cv2.FONT_HERSHEY_PLAIN
 
 
-def put_text(img: np.ndarray, text: str, position: Tuple[int, int],
-             color: Tuple[int, int, int] = (255, 255, 255)) -> np.ndarray:
+def put_text(
+        img: np.ndarray,
+        text: str,
+        position: Tuple[int, int],
+        font_scale: float = 1,
```
florianletsch, line 17: `1` is not a float. Will be converted of course, but I think `1.` is cleaner

sense/display.py
```diff
@@ -181,8 +191,75 @@ def display(self, img: np.ndarray, display_data: dict) -> np.ndarray:
...
-        put_text(img, "Model FPS: {:.1f}".format(inference_engine_fps), (5, img.shape[0] - self.y_offset), text_color)
+        put_text(img, "Camera FPS: {:.1f}".format(camera_fps), (5, img.shape[0] - self.y_offset - 20),
+                 color=text_color)
+        put_text(img, "Model FPS: {:.1f}".format(inference_engine_fps), (5, img.shape[0] - self.y_offset),
+                 color=text_color)
+
+        return img
+
+
+class DisplayDigits(BaseDisplay):
```
florianletsch, line 202: You are using this for digits, but it could be used for any recognized class name, right? Maybe rename to something more generic (`DisplayClassnameOverlay` or similar)?
Then you can maybe reduce the default font scale so that a typical class name (10 to 20 characters) can be displayed nicely and only your example script sets the larger display parameters.

florianletsch (commented):
Code looking good. I left minor comments.

I've tested it now, but unfortunately, it didn't work really well for me. I feel like our old airdigits demo was way more reliable and I also didn't have to be so careful with the timing of my motion. 
https://drive.google.com/file/d/1kZvOM1iCSyM3fq3alcK8jcNK5Ya8xbOP/view?usp=sharing

scripts/run_digit_recognition.py
```diff
@@ -0,0 +1,80 @@
...
+from sense.controller import Controller
+from sense.downstream_tasks.digit_recognition import INT2LAB, LAB2THRESHOLD
+from sense.downstream_tasks.nn_utils import Pipe, LogisticRegression
+from sense.downstream_tasks.postprocess import PostprocessClassificationOutput
+
+
+if __name__ == "__main__":
+    # Parse arguments
+    args = docopt(__doc__)
+    camera_id = args['--camera_id'] or 0
```
florianletsch, line 34:
```suggestion
    camera_id = int(args['--camera_id']) or 0
```
Just found out why it wasn't working for me yesterday. With an explicit camera id passed in, this becomes a string and OpenCV doesn't digest it properly. I think we have to fix this across all scripts.

sense/display.py
corneliusboehm: Good idea :slightly_smiling_face:

scripts/run_digit_recognition.py
```diff
@@ -0,0 +1,80 @@
...
+from sense.controller import Controller
+from sense.downstream_tasks.digit_recognition import INT2LAB, LAB2THRESHOLD
+from sense.downstream_tasks.nn_utils import Pipe, LogisticRegression
+from sense.downstream_tasks.postprocess import PostprocessClassificationOutput
+
+
+if __name__ == "__main__":
+    # Parse arguments
+    args = docopt(__doc__)
+    camera_id = int(args['--camera_id']) or 0
```
sunny-panchal, line 34: TypeError for the int() operation if camera id isn't passed in

florianletsch, line 34: My bad...
```suggestion
    camera_id = int(args['--camera_id'] or 0)
```

PR #69, opened 2020-12-18
[Annotation] Fix: data-path error

manikd31:
### Developer Notes
This PR makes the code more consistent and robust to handle data-path as a command-line argument. Relative data-path couldn't successfully fetch the frames generated while annotating, which is not always stored as the complete path, irrespective of what the user inputs:
`--data_path=dataset` or `--data_path=/home/user/.../sense/dataset` will both be treated as `--data_path=/home/user/.../sense/dataset`

#### Additional changes:
- reformat code
- clean-up imports
- translate comments from French to English

**Note:** Need docstrings for Flask methods in `annotation.py`.

### DIFF
--- a/scripts/tools/annotation/annotation.py
+++ b/scripts/tools/annotation/annotation.py
@@ -162,33 +162,6 @@ def train_lr():
     return redirect(url_for('annot', nom=num))
 
 
-# @app.route('/export_annotation', methods=['POST'])
-# def export_annotation():
-#     data = request.form       # a multi-dict containing POST data
-#     num = int(data['num'])
-#     new_annotations = []
-#     for label, tags in export_labels.items():
-#         annotations_tagged = glob.glob(join('annotations', label, 'tags', '*', 'annotation.json'))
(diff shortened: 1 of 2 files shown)


scripts/tools/annotation/annotation.py
```diff
@@ -8,33 +8,38 @@
...
+from flask import send_from_directory
+from flask import url_for
+from joblib import dump
+from joblib import load
+from os.path import join
+from sklearn.linear_model import LogisticRegression
 
 args = docopt(__doc__)
-dataset_path = args['--data_path']
+dataset_path = join(os.getcwd(), args['--data_path'])
```
antoinemrcr, line 34: will it work if you give an absolute path ?

manikd31: Yes, I did try that and it worked in both cases:
- absolute path
- relative path

`os.getcwd()` and `os.path.join()` overlaps the parent path and that's why this works.

PR #70, opened 2020-12-18
[Fix] flip_video update 

YasheshSavani:
This PR will add a fix to `flip_video.py` for `path_in` and `path_out` variable so it can work with relative and absolute path both.


manikd31:
Makes sense! :+1:

manikd31 (approved):

PR #74, opened 2020-12-22
[Fix] Pathlib replaced with os

YasheshSavani:
This PR will replace Pathlib based code to os based code.


manikd31 (approved):

corneliusboehm:
Just out of curiosity: Why are we going back to `os.path`? I like `pathlib` :grin:

manikd31:
> Just out of curiosity: Why are we going back to `os.path`? I like `pathlib` grin

@antoinemrcr suggested we shouldn't use `pathlib` and stick to `os.path` as we do in the other scripts as well. IMO, I also prefer `pathlib` :stuck_out_tongue:

corneliusboehm:
So an alternative would have been to change the other scripts to `pathlib` as well, right? :grin: 
But if that has been decided already, I'll stay quiet. Most of all, I'm for consistency as well :wink:

antoinemrcr:
I don t like pathlib, i prefer os

corneliusboehm:
Should we call a vote? :P

antoinemrcr (dismissed):
LGTM

sunny-panchal:
> Should we call a vote? :P

I don't have a strong opinion...for consistency we could stick to os for now, but in the long run, why do you prefer os over pathlib @antoinemrcr? And why pathlib over os @corneliusboehm?

scripts/tools/flip_video.py
```diff
@@ -17,29 +17,30 @@
...
 from docopt import docopt
-from pathlib import Path
 
 if __name__ == '__main__':
+    # Parse arguments
     args = docopt(__doc__)
-    videos_path_in = Path.cwd() / Path(args['--path_in'])
-    videos_path_out = Path.cwd() / Path(args['--path_out']) if args.get('--path_out') else videos_path_in
+    videos_path_in = os.path.join(os.getcwd(), args['--path_in'])
+    videos_path_out = os.path.join(os.getcwd(), args['--path_out']) if args.get('--path_out') else videos_path_in
```
sunny-panchal, line 28:
```suggestion
    videos_path_in = os.path.join(os.getcwd(), args['--path_in'])
    videos_path_out = os.path.join(os.getcwd(), args.get('--path_out', args['--path_in']))
```

antoinemrcr:
pathlib is not straight forward to understand, there is a learning curve that make the project harder to understand

scripts/tools/flip_video.py
YasheshSavani: This change is giving NoneType error

manikd31 (approved):
Looks good to me :+1:

sunny-panchal (approved):
works :+1:

PR #75, opened 2021-01-05
Add LogisticRegression with sigmoid outputs

corneliusboehm:
Add support for sigmoid outputs in the `LogisticRegression` class (not used anywhere yet)

### DIFF
--- a/sense/downstream_tasks/nn_utils.py
+++ b/sense/downstream_tasks/nn_utils.py
@@ -72,10 +72,8 @@ class Pipe(RealtimeNeuralNet):
 
 class LogisticRegression(nn.Sequential):
 
-    def __init__(self, num_in, num_out, use_sigmoid=False, use_softmax=True, global_average_pooling=True):
+    def __init__(self, num_in, num_out, use_softmax=True, global_average_pooling=True):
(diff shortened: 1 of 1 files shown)


sunny-panchal:
Is it intended as an activation function before the softmax or would you use either one exclusively?

corneliusboehm:
> Is it intended as an activation function before the softmax or would you use either one exclusively?

Theoretically, I've left it open, but in practice I have only used them exclusively. Do you think we should make sure to only use either or?

sunny-panchal:
> > Is it intended as an activation function before the softmax or would you use either one exclusively?
> 
> Theoretically, I've left it open, but in practice I have only used them exclusively. Do you think we should make sure to only use either or?

Not sure... I don't believe sigmoid would be a good activation before a softmax in general in case of saturation so there may never be a reason to use them together. I was also wondering if it would make sense instead to have a separate sub-class of the logistic regression with the sigmoid output layer instead.

corneliusboehm:
Made it its own subclass now, which also makes sure to disable softmax

corneliusboehm:
Rebased this branch on `master` now so we can already merge it

sunny-panchal (approved):
lgtm :+1:

PR #76, opened 2021-01-06
Fix ID for second label radio button

corneliusboehm:
In the current `master` version, the `1` button isn't clickable, because the ID has a typo.


antoinemrcr (approved):
LGTM

PR #77, opened 2021-01-07
Fix annotation buttons

corneliusboehm:
1. The link for `train_logreg` was broken
2. The "Export annotations" button seems outdated, because the `/export_annotation` endpoint no longer exists. The functionality is now probably covered by the submit button on the main form.


antoinemrcr (approved):
LGTM

PR #80, opened 2021-01-15
Update README.md

manikd31:
This PR adds licensing information for using audio files for audible-alerts during `video-capture`.


sunny-panchal:
@florianletsch should there also be a separate license file where the audio files are located?

florianletsch:
> @florianletsch should there also be a separate license file where the audio files are located?

I think this is fine as is. The CC0 license isn't very strict on how (or even if) the author is mentioned. Keeping it in the README is fine, afaik

florianletsch (approved):

PR #81, opened 2021-01-15
AV-3808: Set up CI/CD

YasheshSavani:
Ticket: https://20bn.atlassian.net/browse/AV-3808

Description:
This PR will add `.travis.yml` file into the repository which will be used for CI/CD on Travis and unit tests for the `display.py` file.


requirements.txt
```diff
@@ -15,4 +15,8 @@ simpleaudio             ==1.0.4       # MIT
 tensorflow              ==1.14.0
 keras                   ==2.2.4
 coremltools             ==3.3
-h5py                    ==2.10.0
\ No newline at end of file
+h5py                    ==2.10.0
+
+# unit tests
+pycodestyle       ==2.5.0          # MIT
+pytest            ==5.4.3          # MIT
```
guillaumebrg, line 22: Let's align this with the rest

sense/downstream_tasks/nn_utils.py
```diff
@@ -1,6 +1,8 @@
+import os
+import torch
```
guillaumebrg, line 2: Let's move `import torch` to just above `import torch.nn as nn`

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+python:
+  - "3.6"
+
+jobs:
+  include:
+    - name: test
+      services:
+        - xvfb
+      before_script:
+        - "export DISPLAY=:99.0"
```
guillaumebrg, line 13: What is this used for? Maybe add a comment. We could explain `xvfb` too.

guillaumebrg, line 13: I'm surprised not to see `export TRAVIS=true` anywhere? Why isn't that needed?

tests/unit/test_display.py
```diff
@@ -0,0 +1,60 @@
+import unittest
```
guillaumebrg, line 1: Since we don't have a folder for additional integration tests yet, let's put this file directly under `tests` remove the intermediate folder `unit` for now.

guillaumebrg (commented):
Great! Posted a bunch of minor comments and a few questions

.travis.yml
YasheshSavani: That is not needed as Travis internally provides an environment variable named `TRAVIS` with value `'true'`. This is we are using in our script.

tests/unit/test_display.py
YasheshSavani: ok

sense/downstream_tasks/nn_utils.py
```diff
@@ -1,6 +1,8 @@
+import os
+
 import numpy as np
+import torch
 import torch.nn as nn
```
sunny-panchal, line 5:
```suggestion
import os
import torch

import numpy as np
import torch.nn as nn
```

sunny-panchal (commented):
Looks good; just left a minor change request for some imports. The tests for display text will break if the text position is changed so we could make that more robust later.

tests/test_demo_scripts.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "Test Fitness Tracker demo ::"
+    demo_fitness_tracker
+
+    echo "Test Gesture Recognition demo ::"
+    demo_gesture_recognition
+}
+
+# Run the created function
+test_demos
+echo "All tests passed!"
```
guillaumebrg, line 37: The script looks a bit overkill to me. How about just this:
```
#!/bin/bash

echo "Test Calorie Estimation demo ::"
PYTHONPATH=./ python scripts/run_calorie_estimation.py --weight=65 --age=30 --height=170 --gender=female --path_in=tests/resources/test_video.mp4

echo "Test Fitness Rep Counter demo ::"
PYTHONPATH=./ python scripts/run_fitness_rep_counter.py --path_in=tests/resources/test_video.mp4

echo "Test Fitness Tracker demo ::"
PYTHONPATH=./ python scripts/run_fitness_tracker.py --weight=65 --age=30 --height=170 --gender=female --path_in=tests/resources/test_video.mp4

echo "Test Gesture Recognition demo ::"
PYTHONPATH=./ python scripts/run_gesture_recognition.py --path_in=tests/resources/test_video.mp4
```

and I wouldn't print "All tests passed!" because I think this will be always be printed, even when if one of the python scripts fails

YasheshSavani: yes, this can be done

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+        # requirement for simpleaudio to work
+        - sudo apt-get install libasound2-dev
+        - pip install -qr requirements.txt
+      script:
+        - pycodestyle
+        - python -m pytest
+        - PYTHONPATH=./ python scripts/run_calorie_estimation.py --weight=65 --age=30 --height=170 --gender=female --path_in=tests/resources/test_video.mp4
+        - PYTHONPATH=./ python scripts/run_fitness_rep_counter.py --path_in=tests/resources/test_video.mp4
+        - PYTHONPATH=./ python scripts/run_fitness_tracker.py --weight=65 --age=30 --height=170 --gender=female --path_in=tests/resources/test_video.mp4
+        - PYTHONPATH=./ python scripts/run_gesture_recognition.py --path_in=tests/resources/test_video.mp4
```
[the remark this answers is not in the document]
YasheshSavani: we can replace them with bash script/

guillaumebrg (approved):
:tada:

PR #82, opened 2021-01-15
Fix files based on pycodestyle rules

YasheshSavani:
This PR will update the files adhering to the `pycodestyle` rules.


sunny-panchal:
Looks good but these changes should be rebased to master and merged there instead

YasheshSavani:
> Looks good but these changes should be rebased to master and merged there instead

Okay. I'll do that

PR #85, opened 2021-01-20
Provide better exception for load weights method

nahuakang:
Just a quick fix :smile:


sense/downstream_tasks/nn_utils.py
```diff
@@ -114,7 +115,15 @@ def load_weights(checkpoint_path: str):
     """
     try:
         return torch.load(checkpoint_path, map_location='cpu')
-    except:
-        raise Exception('ERROR - Weights file missing: {}. To download, please go to '
-                        'https://20bn.com/licensing/sdk/evaluation and follow the '
-                        'instructions.'.format(checkpoint_path))
+    except FileNotFoundError:
+        root_dir = Path(__file__).parent.parent.parent
+        exec_dir = Path(os.path.dirname(os.path.realpath("__main__")))
```
corneliusboehm, line 120: I think recently we have replaced all usages of `Path` with their `os.path` counterparts. Until we've agreed on a project-wide convention, we should probably stick to that :grimacing:

corneliusboehm (commented):
Even though the check for `.parent.parent.parent` is very specific, I like the new error message a lot, because it's probably a very common error source :slightly_smiling_face:

sense/downstream_tasks/nn_utils.py
corneliusboehm: Idea for the `.parent.parent.parent` issue: Could we walk up the directory tree until we find the `sense` package?

nahuakang: I was hesitant on the `.parent.parent.parent` as well. One way I can think of is to have a `utils.py` file in a fixed location in the project where we can store a `ROOT_DIR` const or a function.

We could also walk up the directory tree until sense, but how? StackOverflow did not answer this question on my quick search :laughing: 

Your second thought is very interesting! Let's see how @guillaumebrg thinks as well.

corneliusboehm: > We could also walk up the directory tree until sense, but how? StackOverflow did not answer this question on my quick search laughing

The code snippet above is how :wink:

guillaumebrg: @nahuakang @corneliusboehm, I was about to make a small change to the PR and ended up getting carried away... :) 
My proposal is here: https://github.com/TwentyBN/sense/pull/90
Let me know what you think

guillaumebrg (approved):
Alright... let's merge this :+1:

PR #86, opened 2021-01-22
SenseStudio

corneliusboehm:
Adding SenseStudio as the tool to setup and maintain all of your projects. The annotation tool is integrated.

A structural change is that the previous `scripts` directory was split into `examples`, containing all demo scripts (gesture recognition etc.), and `tools`, containing the training and recording scripts as well as SenseStudio. This way the depth of the folder structure was reduced by one.

There are still a few open issues, but the current version should support everything needed in the beginning and be relatively stable :crossed_fingers:


antoinemrcr (approved):
LGTM

PR #87, opened 2021-01-22
Tensorflow lite

antoinemrcr:
migrate base conversion for iOS from coreml to tensorflow-lite


sunny-panchal:
Successfully converted a model and built the project on iOS with no issues; model was confirmed to work :+1: 

I'd suggest moving the instruction to install pod to the `requirements` section of the readme. Also, it was `brew install cocoapods` that I had to run

requirements.txt
```diff
@@ -11,12 +11,16 @@ Pillow                  ==8.0.1       # HPND
...
 simpleaudio             ==1.0.4       # MIT
 
-# Dependencies needed for converting Pytorch models to CoreML
-tensorflow              ==1.14.0
-keras                   ==2.2.4
-coremltools             ==3.3
-h5py                    ==2.10.0
+# Dependencies needed for converting Pytorch models to tensorflow Lite
+tensorflow              ==2.3.1
+Keras                   ==2.4.3
```
florianletsch, line 16: I know the file wasn't complete with licenses before, but can we agree to add license information here from now on? Both tensorflow and Keras are: `Apache 2.0`

requirements.txt
```diff
@@ -11,12 +11,16 @@ Pillow                  ==8.0.1       # HPND
...
 pycodestyle             ==2.5.0          # MIT
-pytest                  ==5.4.3          # MIT
\ No newline at end of file
+pytest                  ==5.4.3          # MIT
+
+# Dependencies needed for converting Pytorch models to CoreML
+#tensorflow              ==1.14.0
+#keras                   ==2.2.4
+#coremltools             ==3.3
+#h5py                    ==2.10.0
```
florianletsch, line 26: Commented out? Are these lines leftover from development and no longer needed?

tools/conversion/convert_to_tflite.py
```diff
@@ -0,0 +1,1185 @@
...
+            outputs = []
+
+            for f in range(inputs_needed):
+                #                print('input index: ', len(all_layers) - inputs_needed + f, ' shape: ',
+                #                      all_layers[len(all_layers) - inputs_needed + f].shape)
+                inputs.append(all_layers[len(all_layers) - inputs_needed + f])
+                if merge_in > 0:
+                    #                    print('merge input index: ', len(all_layers) - inputs_needed + f, ' shape: ',
+                    #                          all_layers[len(all_layers) - (2 * inputs_needed) + f].shape)
+                    inputs.append(all_layers[len(all_layers) - (2 * inputs_needed) + f])
```
florianletsch, line 245: This script is so messy :( Will we ever clean it up?

requirements.txt
antoinemrcr: only needed if you want to a conversion to coreml. (explained in the comment)
conversion in tensorflow lite or coreml have different dependencies versions

tools/conversion/convert_to_tflite.py
antoinemrcr: Not planned in this PR.

requirements.txt
florianletsch: Ah, got it. Not very easy to handle for users, though, I think (and also not testable). How about we move these dependencies to a different requirements file that can be installed in addition to the normal requirements? For example, call it `requirements-coreml.txt`?

guillaumebrg (approved):
Let's clean these two conversion scripts soon... I'm adding tickets to our backlog

PR #89, opened 2021-01-22
[Fix] AV-4206: Make window resizable

YasheshSavani:
Description:
This PR will make the display window resizable by the user. It will also pad videos on the sides if the videos have resolution low in compare to the default resolution which is `width: 480` and `height: 640`.


guillaumebrg:
Just tested it and the window is resizeable now.

> It will also pad videos on the sides if the videos have resolution low in compare to the default resolution which is width: 480 and height: 640

Is that necessary? Why do we want this?

YasheshSavani:
> Just tested it and the window is resizeable now.
> 
> > It will also pad videos on the sides if the videos have resolution low in compare to the default resolution which is width: 480 and height: 640
> 
> Is that necessary? Why do we want this?

@Yujie was facing this issue. In the top border, the probability output was not shown properly it was getting cut because her video stream was of size `440x440` which is lower than expected `640x480` to show output's properly on the black border. Right now, our offsets for displaying the data are based on the default expected resolution `640x480`.

guillaumebrg:
Got it! Thanks for explaining

Let's maybe add a `window_size` argument to `DisplayResults` then, and I would suggest to unit test this ;)

guillaumebrg (approved):
LGTM! :+1:

PR #90, opened 2021-01-22
Load weights from resources folder

guillaumebrg:
Note: this branch includes commits from #85. 

Rather than making the warning message smarter, this PR should make it possible to run our examples from other places than the root dir (as long as `sense` is in the PYTHONPATH and can be imported)

Let me know what you think


nahuakang (approved):
Looks good to me! This is how I thought we could approach the root dir as well. Unfortunately it seems that `Path` is still not making its way into the repo :laughing:

guillaumebrg:
Thanks for the review @nahuakang!

> Unfortunately it seems that Path is still not making its way into the repo

We could switch to using `Path` but, if we do, I think we should use it consistently throughout the entire repo.

PR #91, opened 2021-01-26
Allow adding new classes on the project details page

corneliusboehm:
New classes can now be added on the project details page through this interface:
image

Safety checks so that a unique class name has to be provided were added.
If no tag names are provided, default names are created.


tools/sense_studio/sense_studio.py
```diff
@@ -255,6 +272,38 @@ def project_details(path):
...
+    path = projects[project]['path']
+
+    # Get class name and tags
+    data = request.form
+    class_name = data['class']
+    tag1 = data['tag1'] or f'{class_name}_tag1'
+    tag2 = data['tag2'] or f'{class_name}_tag2'
+
+    if tag2 == tag1:
+        tag2 = f'{tag2}_2'
```
nahuakang, line 291: If I input `test5_1` as the tag name for both tags, then the second tag would be renamed to `test5_1_2`. But wouldn't it then make sense to rename the first one as `test5_1_1` as well? :thinking:

nahuakang (dismissed):
:jack_o_lantern: Looking good! Tested the new feature against edge cases and it works. Only one question below.

tools/sense_studio/sense_studio.py
corneliusboehm: We could do that :+1:

nahuakang (approved):
Methinks it's time for approval :apple:

PR #92, opened 2021-01-26
Drop tensorflow 1

florianletsch:
The CoreML copnversion script requires Tensorflow 1 to run while the TFLite conversion scripts requires Tensorflow 2.

This PR drops the TF1 dependency and changes the coreml conversion to work with Tensorflow 2 instead.

Still need to confirm that the output model works correctly in iOS before this can merged.

Another open question: We still include keras from the `keras` package - but it's now completely part of `tensorflow`, so I think we can simply fix the Python imports and not install `keras` as an extra package.

### DIFF
--- a/requirements-coreml.txt
+++ b/requirements-coreml.txt
@@ -1,5 +1,0 @@
-# package                version       license
-tensorflow              ==1.14.0       # Apache 2.0
-keras                   ==2.2.4        # Apache 2.0
-coremltools             ==3.3          # BSD 3
-h5py                    ==2.10.0       # BSD 3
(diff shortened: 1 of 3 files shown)


guillaumebrg:
@florianletsch While you're at it, can you check if we could switch to `tensorflow>=2.4.0`? 
We currently specify 2.3.1 but this version of tensorflow has vulnerabilities that triggered Dependabot alerts (I dismissed them because these didn't some important for our use of tensorflow)

antoinemrcr:
Can we close this PR ?

florianletsch:
Yes, it's obsolete.

PR #93, opened 2021-01-26
continuation of checkpoints: custom classifiers loading issue for training

yoga-0125:
Allow training continuation from a checkpoint (AV-3991):

Files update:

- tools/train_classifier.py: add options --save_checkpoints, --frequency=NUM, --load_checkpoints=PATH
- sense/finetuning.py: update training_loops function for updating last_classifier, loading / saving epoch and optimizer
- tools/run_custom_classifier.py: update custom classifier loading (name, location, key)
- add sense/utils.py to save functions (avoid circular dependency) 

Unsolved problems:

- not sure which information related to loss need to be saved and loaded for resume training
- loss (train/val) increase and top 1 (train/val) decrease when start resume training (show in figures)


tools/train_classifier.py
```diff
@@ -10,6 +10,10 @@
                        [--path_annotations_train=PATH]
                        [--path_annotations_valid=PATH]
                        [--temporal_training]
+                       [--save_checkpoints]
```
sunny-panchal, line 13: I think this should be true by default

tools/train_classifier.py
```diff
@@ -62,11 +63,41 @@ def clean_pipe_state_dict_key(key):
...
+    resume_epoch = -1
+    if weights:
+        # Load custom and original classifiers
+        checkpoint_classifier = load_weights(weights)
+        checkpoint = torch.load('resources/backbone/strided_inflated_efficientnet.ckpt')
+        # Update original weights in case some intermediate layers have been finetuned
+        name_finetuned_layers = set(checkpoint.keys()).intersection(checkpoint_classifier.keys())
+        for key in name_finetuned_layers:
+            checkpoint[key] = checkpoint_classifier.pop(key)
+        feature_extractor.load_state_dict(checkpoint['model_state_dict'])
```
sunny-panchal, line 88: As we discussed, the model state dict should be loaded on `net`

tools/train_classifier.py
```diff
@@ -62,11 +63,41 @@ def clean_pipe_state_dict_key(key):
...
+    elif resume:
+        checkpoint_classifier = load_weights(os.path.join(path_out, 'checkpoints/', 'last_classifier.checkpoint'))
+        checkpoint = torch.load('resources/backbone/strided_inflated_efficientnet.ckpt')
+        name_finetuned_layers = set(checkpoint.keys()).intersection(checkpoint_classifier.keys())
+        for key in name_finetuned_layers:
+            checkpoint[key] = checkpoint_classifier.pop(key)
+        resume_epoch = checkpoint['epoch']
```
sunny-panchal, line 95: `checkpoint` here is the loaded backbone weights, not the saved checkpoint file from training, hence the `KeyError`

sunny-panchal (commented):
Hey Yujie, just a summary of what we discussed:
- The checkpoint file would apply to `net` (Piped model with the fine tuned layers and gesture classifier) and so this is where you'd have to load the weights and training parameters to. 
- `--resume` and `--weights` can be combined as `--load_checkpoint` which can have a default value of `-1` that loads the latest checkpoint file
- Additionally save the `num_fine_tuned_layers` in the checkpoint and for now, use that value instead in case the user passes in the parameter again -- we can update this later
- For resuming training from a checkpoint, the optimizer state dict and loss would additionally be needed in the checkpoint file: https://pytorch.org/tutorials/beginner/saving_loading_models.html

I've left some other comments on the code itself. Also, you mentioned it was saving `.zip` files for the checkpoints for you but it was saving `.checkpoints` for me...

sense/finetuning.py
```diff
@@ -268,21 +270,41 @@ def extract_features(path_in, net, num_layers_finetune, use_gpu, num_timesteps=1
...
+    for epoch in range(resume_epoch + 1, num_epochs + resume_epoch + 1):  # loop over the dataset multiple times
         new_lr = lr_schedule.get(epoch)
         if new_lr:
             print(f"update lr to {new_lr}")
             for param_group in optimizer.param_groups:
                 param_group['lr'] = new_lr
 
+        # rename the last saved checkpoint
+        if epoch % frequency == 1 and \
+                os.path.exists(os.path.join(path_out, "checkpoints/", "classifier%g.checkpoint" % (epoch - 1))):
```
antoinemrcr, line 304: can we use a f-string here ? would make it easier to read

antoinemrcr:
Do we need to save checkpoints every n epochs ?
Can t we just save the last one every epoch, and the best model if a better validation loss is found ?

sense/finetuning.py
yoga-0125: > can we use a f-string here ? would make it easier to read

yes, I'll try use f-string format here, thanks

yoga-0125:
> Do we need to save checkpoints every n epochs ?
> Can t we just save the last one every epoch, and the best model if a better validation loss is found ?

I'm not sure, just see some other algorithms (e.g. yolov3) save checkpoints every n epochs. users could pick one specific checkpoint to restart running instead of resume using last_classifier

antoinemrcr:
> > Do we need to save checkpoints every n epochs ?
> > Can t we just save the last one every epoch, and the best model if a better validation loss is found ?
> 
> I'm not sure, just see some other algorithms (e.g. yolov3) save checkpoints every n epochs. users could pick one specific checkpoint to restart running instead of resume using last_classifier

If you want to keep this feature, i would not make it a default.
I would like if by default we don t save every n epochs, but just the last one.

yoga-0125:
> > > Do we need to save checkpoints every n epochs ?
> > > Can t we just save the last one every epoch, and the best model if a better validation loss is found ?
> > 
> > 
> > I'm not sure, just see some other algorithms (e.g. yolov3) save checkpoints every n epochs. users could pick one specific checkpoint to restart running instead of resume using last_classifier
> 
> If you want to keep this feature, i would not make it a default.
> I would like if by default we don t save every n epochs, but just the last one.

Sure, i'll update this part, thanks. And how about the loss problem? any idea?

antoinemrcr:
I would simplify the PR :
- don t do a check on the number of layer you want to finetune. Instead, do a check if a training would overwrite any checkpoints / label file.
- In a first version, let s just save the weights, no need to save the optimizer state and number of epochs i Think.
- Save the last checkpoint every epochs. and not just copy the file after n epochs (if you want to save every 50 epochs, chances are that your last checkpoint will not be the last one)

yoga-0125:
> I would simplify the PR :
> 
> * don t do a check on the number of layer you want to finetune. Instead, do a check if a training would overwrite any checkpoints / label file.
> * In a first version, let s just save the weights, no need to save the optimizer state and number of epochs i Think.
> * Save the last checkpoint every epochs. and not just copy the file after n epochs (if you want to save every 50 epochs, chances are that your last checkpoint will not be the last one)

I'll do some changes and give you update, thank you. one more thing I concerned is the learning scheduler (lr_schedule = {0: 0.0001, 40: 0.00001}). If the program has learned more than 40 epochs (lr reduces from 0.0001 to 0.00001), and system breaks. Then it resumes at epoch 0 with learning rate 0.0001, it may affect on the final result.

sense/finetuning.py
```diff
@@ -305,6 +306,15 @@ def training_loops(net, train_loader, valid_loader, use_gpu, num_epochs, lr_sche
                 best_loss = valid_loss
                 best_state_dict = net.state_dict().copy()
 
+        # save the last checkpoint
+        model_state_dict = net.state_dict().copy()
+        model_state_dict = {clean_pipe_state_dict_key(key): value
+                            for key, value in model_state_dict.items()}
+        last_classifier = os.path.join(path_out, "checkpoints/", "last_classifier.checkpoint")
+        torch.save({
+            'model_state_dict': model_state_dict,
```
antoinemrcr, line 315: you can directly do a torch.save(model_state_dict)
then We don t need to do: checkpoint_classifier['model_state_dict']

yoga-0125: > you can directly do a torch.save(model_state_dict)
> then We don t need to do: checkpoint_classifier['model_state_dict']

yes, I'll update it soon.

tools/train_classifier.py
```diff
@@ -62,15 +56,40 @@ def clean_pipe_state_dict_key(key):
...
+
+    # Create a folder "checkpoints" to save model weights
+    path = os.path.join(path_in, "checkpoints")
+    os.makedirs(path, exist_ok=True)
 
     # Load feature extractor
     feature_extractor = feature_extractors.StridedInflatedEfficientNet()
-    checkpoint = torch.load('resources/backbone/strided_inflated_efficientnet.ckpt')
+
+    if resume:
```
antoinemrcr, line 68: i think you should not make the check on resume.
Check if the different files that will be created with the training ( label2int, checkpoints, ...) are already existing in the output folder

yoga-0125: > i think you should not make the check on resume.
> Check if the different files that will be created with the training ( label2int, checkpoints, ...) are already existing in the output folder

I'll move the warning outside resume. In addition, is the warning code ok? Or I have to check each file name (xxx.checkpoint, xxx.json, xxx.png, xxx.npy)?

antoinemrcr (approved):
LGTM

PR #94, opened 2021-01-26
Refactor tflite converter

florianletsch:
The converter scripts are ... messy. This is a first attempt to refactor the TFLite converter.


antoinemrcr (approved):
Looks good to me

PR #95, opened 2021-01-26
[hotfix] fix load weights for custom classifier

antoinemrcr:
Fix the weights loading for running custom classifier


guillaumebrg (approved):
Thanks for the fix @antoinemrcr!

I've made a few edits so that this script is covered by our tests. I should have created another PR for this, sorry... I thought I checked out another branch. Anyway, merging this now

PR #97, opened 2021-01-27
Removed "kit" from logo

sunny-panchal:
The logo here shouldn't have "kit" in it. We should decide on how to reference SenseKit in the readme though.


corneliusboehm (approved):
Looks good, sorry for missing it before

PR #99, opened 2021-01-29
Remove CoreML converter

florianletsch:
Removing coreml converter script until it's ported to Tensorflow 2 and introduced in existing converter structure


antoinemrcr (approved):
LGTM

PR #84, opened 2021-01-18
Add classname display overlay

corneliusboehm:
This PR copies over some general changes from #64 as long as that PR is still on hold.

- Add `font_scale` and `thickness` arguments to the `put_text` method
- Add `DisplayClassnameOverlay` operation that shows detected classes as a big text overlay (see screenshot)
- Fix `int` parsing of `camera_id` in all scripts
- Fix assignment of `runtime_error` in `controller.py` (the `as` statement wasn't working as expected)

facepalming

If you want to test the `DisplayClassnameOverlay`, add this to the `display_ops` in `run_gesture_recognition.py`:
```python
        sense.display.DisplayClassnameOverlay(
            thresholds={
                "\"Sleeping\" gesture": 0.5,
                "Covering ears": 0.5,
                "Covering eyes": 0.5,
                "Facepalming": 0.5,
                "Nodding": 0.5,
                "Putting finger to mouth": 0.5,
                "Scratching": 0.5,
                "Shaking head": 0.5,
                "Waving": 0.5,
            }
        )
```
We could even add this to the script per default if people like it.

PR #101, opened 2021-02-01
Display class name on gesture recognition demo

nahuakang:
This PR is a follow-up ticket to [PR84](https://github.com/TwentyBN/sense/pull/84) and uses `DisplayClassNameOverlay` in the `run_gesture_recognition.py` script.

The PR also introduces a new private method: `DisplayClassnameOverlay._adjust_font_scale` to automatically re-scale the `font_scale` instance variable should the text width be larger than the frame width.


examples/run_gesture_recognition.py
```diff
@@ -59,6 +60,7 @@
         sense.display.DisplayFPS(expected_camera_fps=net.fps,
                                  expected_inference_fps=net.fps / net.step_size),
         sense.display.DisplayTopKClassificationOutputs(top_k=1, threshold=0.5),
+        sense.display.DisplayClassnameOverlay(thresholds=LAB_THRESHOLDS),
```
corneliusboehm, line 63: The `border_size` used in `DisplayResults` is different depending on if a `title` is given. Therefore, the text isn't currently always perfectly centered.

sense/downstream_tasks/gesture_recognition/__init__.py
```diff
@@ -32,3 +32,33 @@
...
+    "Pointing to the camera": .5,
+    "Putting finger to mouth": .5,
+    "Rolling hand": .5,
+    "Scratching": .5,
+    "Shaking head": .5,
+    "Showing the middle finger": .3,
+    "Swiping down": .4,
+    "Swiping down (with two hands)": .5,
+    "Swiping left": .5,
+    "Swiping right": .5,
```
corneliusboehm, line 56: Note: Swiping left and right is also mixed up (same as pointing left/right)

sense/downstream_tasks/gesture_recognition/__init__.py
```diff
@@ -32,3 +32,33 @@
...
+    "Facepalming": .5,
+    "Nodding": .5,
+    "Pointing left": .3,
+    "Pointing right": .3,
+    "Pointing to the camera": .5,
+    "Putting finger to mouth": .5,
+    "Rolling hand": .5,
+    "Scratching": .5,
+    "Shaking head": .5,
+    "Showing the middle finger": .3,
```
corneliusboehm, line 52: I wasn't able to activate this at all :shrug: Does it work for you?

sense/downstream_tasks/gesture_recognition/__init__.py
```diff
@@ -32,3 +32,33 @@
...
+    "Swiping down": .4,
+    "Swiping down (with two hands)": .5,
+    "Swiping left": .5,
+    "Swiping right": .5,
+    "Swiping up": .5,
```
corneliusboehm, line 57: Swiping down often triggered immediately after swiping up, because I just lowered my hand again

corneliusboehm (commented):
Nice tuning and testing already :+1:

sense/downstream_tasks/gesture_recognition/__init__.py
nahuakang: It worked for me but I had to lower it to .3, should we lower it even further?

corneliusboehm: I wouldn't want to have false positives on this one. If it worked for you, let's leave it there.

sense/downstream_tasks/gesture_recognition/__init__.py
nahuakang: Increase `Swiping down` up to `.7`? I've always had this issue. Is this more of a model not learning from the full action issue or?

corneliusboehm: I think a higher threshold might help. Give it a try

examples/run_gesture_recognition.py
nahuakang: Hmm, any suggestion here? Shall we add title on top first, let `display_op` inherit `DisplayResults`'s border size, and then display information?

corneliusboehm: Maybe take a look at how I did it in `run_digit_recognition`: https://github.com/TwentyBN/sense/pull/64/files
I'm not perfectly happy with it, but it works for now.

sense/downstream_tasks/gesture_recognition/__init__.py
nahuakang: I tried up to .8, which is too high for a single `Swipe down`. Currently at .65 it is still not perfect because the issue you described is still triggered very often :confused:

corneliusboehm: Ok, then that's just how it is

examples/run_gesture_recognition.py
```diff
@@ -56,11 +56,14 @@
         PostprocessClassificationOutput(INT2LAB, smoothing=4)
     ]
 
+    border_size = 50  # Increase border size for showing top 2 predictions
```
corneliusboehm, line 59: Here we are only showing the top-1 prediction, so the default border size is 30

sense/display.py
```diff
@@ -238,7 +238,9 @@ def __init__(
         self._start_time = None
 
     def _get_center_coordinates(self, img: np.ndarray, text: str, font_scale: float) -> Tuple[int, int]:
-        """Calculate and return the center coordinates for the text on the current frame."""
+        """
+        Calculate and return the coordinates of the lower left corner the text on the current frame.
```
corneliusboehm, line 242:
```suggestion
        Calculate and return the coordinates of the lower left corner of the text for centering it on the current frame.
```

tests/test_display.py
```diff
@@ -70,6 +71,20 @@ def setUp(self) -> None:
...
+    def test_display_below_threshold(self, mock_put_text):
+        test_display = base_display.DisplayClassnameOverlay(thresholds={'Dabbing': 0.5})
+        test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.4]]})
+
+        mock_put_text.assert_not_called()
+
+    @patch('sense.display.put_text')
+    def test_display_class_not_in_threshold(self, mock_put_text):
+        test_display = base_display.DisplayClassnameOverlay(thresholds={'Dabbing': 0.5})
+        test_display.display(self.img, {'sorted_predictions': [['Nodding', 0.4]]})
```
corneliusboehm, line 84: To really make clear this is not a low threshold thing
```suggestion
        test_display.display(self.img, {'sorted_predictions': [['Nodding', 1.0]]})
```

tests/test_display.py
```diff
@@ -81,6 +96,23 @@ def test_display_default(self, mock_put_text):
...
 
+    @patch('sense.display.DisplayClassnameOverlay._display_class_name')
+    def test_display_default_within_duration(self, mock_display_class_name):
+        test_display = base_display.DisplayClassnameOverlay(thresholds={'Dabbing': 0.5}, duration=.1)
+        start_time = time.perf_counter()
+        test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.6]]})
+
+        call_count = 0
+        while time.perf_counter() - start_time <= test_display.duration:
+            test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.6]]})
```
corneliusboehm, line 107: The point here is that the inputs don't even need to contain a high prediction for that class and it will still be displayed

tests/test_display.py
```diff
@@ -81,6 +96,23 @@ def test_display_default(self, mock_put_text):
...
                                          thickness=self.thickness)
 
+    @patch('sense.display.DisplayClassnameOverlay._display_class_name')
+    def test_display_default_within_duration(self, mock_display_class_name):
+        test_display = base_display.DisplayClassnameOverlay(thresholds={'Dabbing': 0.5}, duration=.1)
+        start_time = time.perf_counter()
+        test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.6]]})
+
+        call_count = 0
+        while time.perf_counter() - start_time <= test_display.duration:
```
corneliusboehm, line 106: Here you're actually waiting for the time to pass, which slows your tests down by 1 second. A way better approach is to mock `time.perf_counter` and artificially increase the time.
However, mocking `time` is complicated, which is why you probably need to extract the `now = time.perf_counter()` line from `display()` into its own method so you can mock that.

tests/test_display.py
```diff
@@ -81,6 +96,23 @@ def test_display_default(self, mock_put_text):
...
+            test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.6]]})
+            mock_display_class_name.assert_called_with(self.img, 'Dabbing')
+            call_count = mock_display_class_name.call_count
+
+        # Display something that's not in threshold
```
corneliusboehm, line 111: This comment is confusing. Shouldn't it rather be "Display nothing after duration has passed"?

examples/run_gesture_recognition.py
```diff
@@ -55,10 +56,14 @@
         PostprocessClassificationOutput(INT2LAB, smoothing=4)
     ]
 
+    border_size = 30  # Increase border size for top 1 prediction
```
corneliusboehm, line 59:
```suggestion
    border_size = 30
```

sense/display.py
```diff
@@ -271,6 +295,10 @@ def display(self, img: np.ndarray, display_data: dict):
 
         return img
 
+    @staticmethod
+    def _get_current_time() -> float:
+        return time.perf_counter()
```
corneliusboehm, line 300: Add a comment that this was extracted for easier testing

tests/test_display.py
```diff
@@ -59,6 +63,83 @@ def test_display(self, mock_put_text):
...
+                                         thickness=self.thickness)
+
+    @patch('sense.display.DisplayClassnameOverlay._display_class_name')
+    def test_display_default_within_duration(self, mock_display_class_name):
+        test_display = base_display.DisplayClassnameOverlay(thresholds={'Dabbing': 0.5},
+                                                            duration=self.duration)
+
+        # Set start time and call display once
+        test_display._start_time = test_display._get_current_time()
+        test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.51]]})
```
corneliusboehm, line 108: Isn't the `_start_time` automatically set when you call `display`?

tests/test_display.py
```diff
@@ -59,6 +63,83 @@ def test_display(self, mock_put_text):
...
+        mock_display_class_name.assert_called_with(self.img, 'Dabbing')
+        assert mock_display_class_name.call_count == 1
+
+        # Call display once during the duration
+        with patch("sense.display.DisplayClassnameOverlay._get_current_time",
+                   return_value=test_display._start_time + self.half_duration):
+            test_display.display(self.img, {'sorted_predictions': [['Dabbing', 0.51]]})
```
corneliusboehm, line 115: The point of the duration is that the class name will even be displayed when the input is **below** the threshold. In this current test setup, we wouldn't need a duration at all.

tests/test_display.py
```diff
@@ -59,6 +63,83 @@ def test_display(self, mock_put_text):
...
+
+        # Call display post the duration
+        with patch("sense.display.DisplayClassnameOverlay._get_current_time",
+                   return_value=test_display._start_time + self.duration):
+            # Display nothing after duration has passed
+            test_display.display(self.img, {'sorted_predictions': []})
+
+        # Total call counts should remain the same
+        assert mock_display_class_name.call_count == 2
```
corneliusboehm, line 124: Theoretically it's not clear, where your `call_count` has increased. Could have happened after the first or second call.

corneliusboehm (approved):
:name_badge:

PR #102, opened 2021-02-02
Support editing classes

corneliusboehm:
The class and tag names inside a project can now be edited or even deleted.

### Technical Details
- If you press the Edit button, the regular display card will be hidden and the editing card revealed
image
image
- There is no longer one form encompassing the whole list of class cards, but each card is its own form
- Editing a class name will also rename all corresponding directories for that class
- On editing classes, the same constraints as for adding new classes are checked
image
- Deleting a class will for now not delete any data, to prevent losing anything
image

### Other changes
- The Delete button on the projects overview page is now a link instead of a button, so it integrates more nicely with the card around it. The `a` element had to be changed to inherit color though, so that it wouldn't show up blue.
image
- Adding a class can now handle that the directory already exists (so that you can re-add a class that was previously deleted)


tools/sense_studio/sense_studio.py
```diff
@@ -300,7 +302,79 @@ def add_class(project):
...
+
+    # Get new class name and tags
+    data = request.form
+    new_class_name = data['className']
+    new_tag1 = data['tag1'] or f'{new_class_name}_tag1'
+    new_tag2 = data['tag2'] or f'{new_class_name}_tag2'
+
+    if new_tag2 == new_tag1:
+        new_tag1 = f'{new_tag1}_1'
+        new_tag2 = f'{new_tag2}_2'
```
nahuakang, line 332: Just a nit but do we want to dry this since it's almost the same as the block in `add_class`?

nahuakang (dismissed):
Tested again and works fine. One nit :night_with_stars:

tools/sense_studio/sense_studio.py
corneliusboehm: Yes, thanks for calling me out on that :wink:

nahuakang (approved):
Super. C'est tellement élégant et sec! :wink:

PR #103, opened 2021-02-02
Record video

antoinemrcr:
Add video recording capability in the sense studio.
Needs ffmpeg installed for it to work. What do you think, should we enforce it in the readme ?


tools/sense_studio/sense_studio.py
```diff
@@ -337,6 +338,40 @@ def show_video_list(split, label, path):
     return render_template('video_list.html', folders=folder_id, split=split, label=label, path=path)
 
 
+@app.route('/record_video/<split>/<label>/<path:path>')
+def record_video(split, label, path):
+    """
+    Show the list of videos for the given split, class label and project.
+    If the necessary files for annotation haven't been prepared yet, this is done now.
```
corneliusboehm, line 345: Copy-pasted docstring outdated

tools/sense_studio/sense_studio.py
```diff
@@ -337,6 +338,40 @@ def show_video_list(split, label, path):
...
+    path = f'/{urllib.parse.unquote(path)}'  # Make path absolute
+    if request.method == 'POST':
+        file = request.files['video']
+        temp_file = os.path.join(path, "temp_video.webm")
+        with open(temp_file, "wb") as vid:
+            video_stream = file.read()
+            vid.write(video_stream)
+        output_path = os.path.join(path, "videos_" + split, label)
+
+        os.path.join(output_path, f"video_0.mp4")
```
corneliusboehm, line 364: This line isn't doing anything

tools/sense_studio/sense_studio.py
```diff
@@ -337,6 +338,40 @@ def show_video_list(split, label, path):
...
+        # find a video name that is not used yet
+        for i in range(100000):
+            output_file = os.path.join(output_path, f"video_{i}.mp4")
+            if not os.path.isfile(output_file):
+                break
```
corneliusboehm, line 369: This could maybe be done more efficiently with a `glob` (not requiring that many `isfile` checks)

tools/sense_studio/sense_studio.py
```diff
@@ -337,6 +338,40 @@ def show_video_list(split, label, path):
...
+    """
+    path = f'/{urllib.parse.unquote(path)}'  # Make path absolute
+    split = urllib.parse.unquote((split))
+    label = urllib.parse.unquote(label)
+    return render_template('record_video.html', split=split, label=label, path=path)
+
+
+@app.route('/video_saving/<split>/<label>/<path:path>', methods=['POST'])
+def save_video(path, label, split):
+    path = f'/{urllib.parse.unquote(path)}'  # Make path absolute
```
corneliusboehm, line 355: `label` and `split` should also be unquoted

tools/sense_studio/templates/project_details.html
```diff
@@ -57,6 +57,10 @@ <h4>Annotated Videos</h4>
                             <i class="edit icon"></i>
                             Annotate
                         </a>
+                        <a class="ui icon button" href="{{ url_for('record_video', split='train', label=class, path=path) }}">
+                            <i class="edit icon"></i>
```
corneliusboehm, line 61: Should get a different icon. Also the two buttons don't fit that well next to each other.
image

tools/sense_studio/static/main.css
```diff
@@ -41,9 +41,26 @@ a {
     width: 100%;
 }
 
+.container { position:relative; }
```
corneliusboehm, line 44: Semantic UI already has a `container` class. Did you mean to overwrite that?

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
+{% extends 'skeleton.html' %}
+
+{% block title %} Record video {% endblock %}
+
+{% block main %}
+<div id="main" class="ui main text container">
+    <h1 class="ui header">
+        Video List
```
corneliusboehm, line 8: Outdated title

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
...
+        </div>
+    </h1>
+    <div>
+        <a>Pre recording duration</a>
+        <input name="duration_selection" type="text"  id="pre_recording_duration" value="3" />
+        <p/>
+        <a>Recording duration</a>
+        <input name="duration_selection" type="text"  id="recording_duration" value="5"/>
+        <p/>
+        <button onclick="record_video()">record</button>
```
corneliusboehm, line 24: Old style inputs and button. Please use Semantic UI elements.
image

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
...
+        <p/>
+        <div class="container" width="640" height="480" >
+        <video id="player" width="640" height="480" controls autoplay>
+            <source src="video/Motion.mp4" type="video/mp4" />
```
corneliusboehm, line 28: Is `video/Motion.mp4` referencing a local file on your system? For me this results in an error message shown until I start recording:
image

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
...
+			  const url = "{{ url_for("save_video", path=path, label=label, split=split) }}".replace(/([^:]\/)\/+/g, "$1");
+			  var blob = new Blob(recordedChunks, {
+				type: "video/webm"
+			  });
+			    const formData = new FormData();
+                formData.append("video", blob);
+                fetch(url, {
+                  method: 'POST',
+                  body: formData,
+                }).then(res => {
```
corneliusboehm, line 80: What is going on with the indentation in this script? Sometimes 2, sometimes 4 spaces, some code blocks on a completely different indentation level for no reason

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
...
+                }).catch(err => {
+                  alert(err);
+                });
+			}
+
+			pre_recording = parseFloat(document.getElementById("pre_recording_duration").value)
+			recording = parseFloat(document.getElementById("recording_duration").value)
+
+            var text = "PRE RECORDING: " + pre_recording + "s";
+			document.getElementById("text_overlay").innerHTML = text;
```
corneliusboehm, line 91: Is this text overlay supposed to show a countdown? For me it just shows a permanent `PRE RECORDING: 3s` message.
image

tools/sense_studio/templates/record_video.html
```diff
@@ -0,0 +1,116 @@
...
+
+
+		  var handleSuccess = function(stream) {
+		    var canvas = document.querySelector("canvas");
+
+			// Optional frames per second argument.
+			// var stream = canvas.captureStream(30);
+			var recordedChunks = [];
+
+			console.log(stream);
```
corneliusboehm, line 55: Some console logs remaining

corneliusboehm (changes requested):
The video recording didn't work for me. I saw my video, but the recording never stopped and no file was written. Maybe this error has something to do with it:
image

Tested on Firefox.

tools/sense_studio/templates/record_video.html
antoinemrcr: No, it is not a countdown, just tell you how many time left.
I don t know how to do this countdown in js

tools/sense_studio/templates/project_details.html
antoinemrcr: is there somewhere a list of icons for semantic ui ?

corneliusboehm: Yes, you can easily search here: https://semantic-ui.com/elements/icon.html

tools/sense_studio/templates/record_video.html
corneliusboehm: A countdown would be very helpful though :thinking: Currently it's hard to tell when the video starts and stops

tools/sense_studio/sense_studio.py
antoinemrcr: I am doing it, but what does it change ? It was working before without it

corneliusboehm: This only gets problematic, when your class names contain spaces

tools/sense_studio/static/video_recording.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+
+function setupRecording(stream, url) {
+    let player = document.getElementById('player');
+    player.srcObject = stream;
+
+    countdownDuration = parseInt(document.getElementById('countdown').value)
+    recordingDuration = parseInt(document.getElementById('duration').value)
```
florianletsch, line 78: Bad practice: These are global variables

tools/sense_studio/static/video_recording.js
```diff
@@ -0,0 +1,137 @@
...
+        setTimeout(displayOverlay, seconds * 1000, 'Get Ready: ' + countdown, 'countdown');
+    }
+
+    // Start recording
+    setTimeout(startRecording, countdownDuration * 1000, stream, recordingDuration, url);
+};
+
+
+function startRecording(stream, recordingDuration, url) {
+    mediaRecorder = new MediaRecorder(stream, {mimeType: 'video/webm; codecs=vp8'});
```
florianletsch, line 92: This is a global variable. I think here you're doing it on purpose. Better would be to wrap the media recorder logic in its own [js module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and have the variable only exist in the module scope.

If it _really_ has to be global (which I think it shouldn't), I suggest we follow the convention of attaching it to the window object explicitly like so:

```
window.mediaRecorder = new MediaRecorder(...);
```

tools/sense_studio/static/video_recording.js
```diff
@@ -0,0 +1,137 @@
...
+    const formData = new FormData();
+    formData.append('video', blob);
+    fetch(url, {
+        method: 'POST',
+        body: formData,
+    }).then(res => {
+        if (res.ok) {
+            displayOverlay('Video Saved', 'saved');
+        } else {
+            displayOverlay('Error: ' + res.status, 'error');
```
florianletsch, line 132: Better style is to use a template literal: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

florianletsch (commented):
Some initial comments on the JS code

tools/sense_studio/static/video_recording.js
corneliusboehm: Sorry, missed those :see_no_evil:

tools/sense_studio/static/video_recording.js
corneliusboehm: I think it doesn't even need to be global, because I pass it over to the stopping function below.

tools/sense_studio/static/video_recording.js
corneliusboehm: Nice, there's the f-string I was looking for :grin:

tools/sense_studio/static/video_recording.js
florianletsch: Ah, got it. So now it's a global variable, but you pass it to the stop function, and the parameter name (being the same) shadows the global variable. :exploding_head: 

In any case, what would you think about using proper modules? Not _required_, but a media recorder like this one could lend itself to starting with a proper code structure

corneliusboehm: Wait, after my latest change there should be nothing global left.

Proper code structure sounds good though :grin: I'll definitely add that to the refactoring story.

florianletsch: Oh yeah I commented before I had refreshed the code view, my comment was still for the previous version. It's fine now.

tools/sense_studio/sense_studio.py
```diff
@@ -376,6 +370,49 @@ def show_video_list(split, label, path):
...
+    with open(temp_file_name, 'wb') as temp_file:
+        temp_file.write(input_stream.read())
+
+    # Find a video name that is not used yet
+    existing_files = set(glob.glob(os.path.join(output_path, 'video_[0-9]*.mp4')))
+    video_idx = 0
+    output_file = os.path.join(output_path, f'video_{video_idx}.mp4')
+    while output_file in existing_files:
+        video_idx += 1
+        output_file = os.path.join(output_path, f'video_{video_idx}.mp4')
```
nahuakang, line 405: I was positively surprised about this block of code when I deleted a video in the middle of the index range and found the newly recorded video took over the file name :)

nahuakang (approved):
I tested and everything works quite intuitively. Deleting videos and refilling new videos named in a consecutive range of indices also works :+1: 

One thing for a mid-term todo list: Make the web pages responsive :smile:

florianletsch (changes requested):
Some more comments. I also found the following bug / unexpected behavior:

- You can still click the Record button while it's recording. It starts a second countdown and the number in the overlay jump back and forth between the several countdowns. At the end, multiple videos are written to disk

nahuakang:
@florianletsch Yes! Thanks for reporting that bug. I found that bug too but didn't know which button I clicked and did not reproduce it, so I thought maybe I did something wrong :see_no_evil:

tools/sense_studio/sense_studio.py
antoinemrcr: this might just create an issue :
if you delete a video without remove extracted features and frames, you will get an issue

tools/sense_studio/templates/video_recording.html
```diff
@@ -0,0 +1,62 @@
...
+                <!-- Automatically filled -->
+            </div>
+            <video id="player" width="640" height="480" autoplay></video>
+        </div>
+    </form>
+</div>
+{% endblock %}
+
+{% block scripts %}
+<script src="{{ url_for('static', filename='video_recording.js') }}"></script>
```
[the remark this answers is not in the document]
corneliusboehm: How did you notice the script being cached?
And could this section from `sense_studio.py` have something to do with it?
```python
@app.after_request
def add_header(r):
    """
    Add headers to both force latest IE rendering engine or Chrome Frame,
    and also to cache the rendered page for 10 minutes.
    """
    r.headers["Cache-Control"] = "no-cache, no-store, must-revalidate"
    r.headers["Pragma"] = "no-cache"
    r.headers["Expires"] = "0"
    r.headers['Cache-Control'] = 'public, max-age=0'
    return r
```

tools/sense_studio/templates/video_recording.html
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+        <div class="ui primary icon button" onclick="recordVideo('{{ url_for('save_video', project=project, label=label, split=split) }}');">
+            <i class="circle icon"></i>
+            Record
+        </div>
+        <div class="ui compact placeholder segment video-container" id="videoContainer">
+            <div class="ui compact message hidden" id="textOverlay">
+                <!-- Automatically filled -->
+            </div>
+            <video id="player" width="640" height="480" autoplay></video>
```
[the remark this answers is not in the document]
corneliusboehm: I followed this answer now and it's working fine: https://stackoverflow.com/a/14455993/6611214

tools/sense_studio/sense_studio.py
corneliusboehm: Good point @antoinemrcr. As part of the refactoring ticket I listed to add integrity tests when entering a dataset. Those test should probably also make sure there are no outdated features and frames.

corneliusboehm:
All comments as well as the bug addressed, except for the caching issue. And I have updated the README a bit.

README.md
```diff
@@ -197,57 +197,52 @@ From our experiments, our estimates correlate well with the workout intensity (i
...
-│   │   ├── video7.mp4
-│   │   ├── video8.mp4
-│   │   └── ...
 │   └── ...
 └── project_config.json
 ```
 - Two top-level folders: one for the training data, one for the validation data.
-- One sub-folder for each class with as many videos as you want (but at least one!)
-- Requirement: videos should have a framerate of 16 fps or higher.
+- One sub-folder for each class that you specify
```
florianletsch, line 225:
```suggestion
- One sub-folder for each class that you specify.
```

florianletsch (approved):
Looking good! I tried again locally, and caching the JS hasn't been a problem anymore. Maybe I tested incorrectly yesterday...

corneliusboehm (approved):

PR #107, opened 2021-02-05
[FIX] Frame flip in pre-processing causing left/right confusion

manikd31:
This PR removes the step in the pre-processing phase which was causing a confusion between left and right labels during predictions in `gesture_recognition` and `fitness_tracker`.

### DIFF
--- a/sense/feature_extractors/mobilenet.py
+++ b/sense/feature_extractors/mobilenet.py
@@ -192,7 +192,7 @@ class StridedInflatedMobileNetV2(RealtimeNeuralNet):
         return self.cnn(video)
 
     def preprocess(self, clip):
-        clip = clip[:, :, :, ::-1].copy()
+        clip = clip.copy()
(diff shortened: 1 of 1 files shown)


manikd31:
@guillaumebrg  @antoinemrcr  is this a possible bug in the pre-processing phase? @sunny-panchal  and I found that the display frame only gets flipped once here before making predictions.

sense/feature_extractors/mobilenet.py
```diff
@@ -192,7 +192,7 @@ def forward(self, video):
         return self.cnn(video)
 
     def preprocess(self, clip):
-        clip = clip[:, :, :, ::-1].copy()
```
corneliusboehm, line 195: I think the `copy` is no longer necessary, because it was probably only used here to resolve the `::-1`

guillaumebrg (approved):
Thanks!

PR #110, opened 2021-02-08
TFlite Converter: Add `--verbose` argument to show detailed log messages

florianletsch:
This PR adds a `--verbose` command-line argument to the TFlite conversion script. When verbose logging is enabled, we set the log level accordingly so that all log messages are shown, including `logging.info()`.

In the PR, I also changed all parameters being passed to `logging.info()` to be well-formatted strings, as logging fails if you pass multiple parameters (this is different to `print(..)` which accepts any number of parameters.

If `--verbose` is not set, the only output is the converted model summary with input and output names. Also, there are always a bunch of tensorflow prints that are unaffected by the new `--verbose` flag.

Example call:

```
python tools/conversion/convert_to_tflite.py \
--backbone=efficientnet \
--classifier=efficient_net_gesture_control \
--output_name=model-converted \
--verbose
```

### DIFF
--- a/tools/conversion/convert_to_tflite.py
+++ b/tools/conversion/convert_to_tflite.py
@@ -119,7 +119,7 @@ def convert(backbone_settings, classifier_settings, output_name, plot_model, ver
     logging.info(f"input_names {in_names}")
     logging.info(f"output_names {out_names}")
     logging.info(f"image_input_names {image_inputs}")
-    logging.info(type(keras_file))
+    logging.info(f"keras file type: {type(keras_file)}")
 
     export_keras_to_tflite(keras_file, tflite_file)
 
--- a/tools/conversion/convert_to_tflite.py
+++ b/tools/conversion/convert_to_tflite.py
@@ -139,8 +139,6 @@ if __name__ == "__main__":
     path_in = args["--path_in"]
     plot_model = args["--plot_model"]
     verbose = args["--verbose"]
-    print(verbose)
-    print("\n\n\n\n\n")
 
     backbone_settings = SUPPORTED_BACKBONE_CONVERSIONS.get(backbone_name)
     if not backbone_settings:
@@ -175,4 +173,5 @@ if __name__ == "__main__":
         classifier_settings,
         output_name,
         plot_model,
-        verbose=verbose)
+        verbose=verbose
+    )
--- a/tools/conversion/convert_to_tflite.py
+++ b/tools/conversion/convert_to_tflite.py
@@ -9,6 +9,7 @@ Usage:
                        [--path_in=PATH]
                        [--plot_model]
                        [--float32]
+                       [--verbose]
   convert_to_tflite.py (-h | --help)
 
 Options:
@@ -19,6 +20,8 @@ Options:
   --plot_model        Plot intermediate Keras model and save as image.
   --float32           Use full precision. By default, this script quantizes weights
                       to 16-bit precision.
+  --verbose           Enable detailed logging
+
   -h --help
 """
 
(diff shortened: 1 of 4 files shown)


tools/conversion/convert_to_tflite.py
```diff
@@ -132,6 +138,9 @@ def convert(backbone_settings, classifier_settings, output_name, plot_model):
     float32 = args["--float32"]
     path_in = args["--path_in"]
     plot_model = args["--plot_model"]
+    verbose = args["--verbose"]
+    print(verbose)
```
sunny-panchal, line 142:
```suggestion
    print(f"Verbose log: {verbose})
```

tools/conversion/convert_to_tflite.py
```diff
@@ -161,4 +170,9 @@ def convert(backbone_settings, classifier_settings, output_name, plot_model):
             "{}".format(classifier_settings["corresponding_backbone"])
         )
 
-    convert(backbone_settings, classifier_settings, output_name, plot_model)
+    convert(
+        backbone_settings,
+        classifier_settings,
+        output_name,
+        plot_model,
+        verbose=verbose)
```
sunny-panchal, line 178:
```suggestion
    convert(
        backbone_settings,
        classifier_settings,
        output_name,
        plot_model,
        verbose=verbose
    )
```

tools/conversion/convert_to_tflite.py
```diff
@@ -110,18 +116,18 @@ def convert(backbone_settings, classifier_settings, output_name, plot_model):
         plot(model, to_file=to_file, show_shapes=True)
         logging.info("Saved model plot to {}".format(to_file))
 
-    logging.info("input_names", in_names)
-    logging.info("output_names", out_names)
-    logging.info("image_input_names", image_inputs)
+    logging.info(f"input_names {in_names}")
+    logging.info(f"output_names {out_names}")
+    logging.info(f"image_input_names {image_inputs}")
     logging.info(type(keras_file))
```
sunny-panchal, line 122:
```suggestion
    logging.info(f"keras file type: {type(keras_file)}")
```

tools/conversion/keras_converter.py
```diff
@@ -99,7 +99,7 @@ def create_keras_model(self):
             outputs=[self.container.all_layers[i] for i in self.container.out_index],
         )
         logging.info("done assembling model")
-        logging.info(model.summary())
+        model.summary(print_fn=logging.info)
 
         # logging.info() all inputs, formatted for use with coremltools Keras convertor
         logging.info("input_names=[")
```
sunny-panchal, line 105: Can we indent the ones inside the loop? With each of these on a new line it's a bit hard to see where it starts/ends:

```
INFO:root:input_names=[
INFO:root:'new_frame',
INFO:root:'frame_back_1',
INFO:root:'frame_back_2',
INFO:root:'frame_back_3',
INFO:root:'cnn.3_history_0',
INFO:root:'cnn.3_history_1',
INFO:root:'cnn.7_history_0',
INFO:root:'cnn.11_history_0',
INFO:root:'cnn.11_history_1',
INFO:root:'cnn.14_history_0',
INFO:root:'cnn.17_history_0',
INFO:root:'cnn.17_history_1',
INFO:root:'cnn.20_history_0',
INFO:root:'cnn.20_history_1',
INFO:root:'cnn.23_history_0',
INFO:root:'cnn.23_history_1',
INFO:root:'cnn.25_history_0',
INFO:root:'cnn.25_history_1',
INFO:root:],
```

sunny-panchal (commented):
A few small details:

tools/conversion/convert_to_tflite.py
florianletsch: Oops, I actually didn't intend to commit that line :sweat_smile:

tools/conversion/keras_converter.py
florianletsch: Thanks. I've added the indentation and actually reversed these lines to use `print()` and not `logging.info()`. I think these lines can be considered actual program output that should be displayed _always_, no matter if `--verbose` is set or not.

nahuakang (approved):
LGTM :deciduous_tree:

PR #111, opened 2021-02-09
README - Edit note about pytorch installation

guillaumebrg:
Problem: the README makes it look like `conda install pytorch torchvision cudatoolkit=10.2 -c pytorch` is a required installation step. 

A few `sense-iOS` users reported getting confused about this during our last hackathon.  My suggestion is to include the example install command into the `Note:` body to make it clear that this step is not mandatory.


sunny-panchal (approved):
Looks good :ok_hand:

PR #112, opened 2021-02-11
Update logos

sunny-panchal:
Logos are updated to the latest versions.

On SenseStudio:
image

Homepage:
image


manikd31:
The logo in dark-mode isn't clearly visible, may need to look into that.
Screenshot from 2021-02-11 11-02-55

sunny-panchal:
A white background had to be added for the logo to display properly with a dark theme. There's currently no way of detecting what someone's github theme setting is (https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981) but once there is, will update it to swap the font colour accordingly.

tools/sense_studio/static/sense_studio_logo.svg
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
```
corneliusboehm, line 1: In SenseStudio I would like to use a logo without background, because for example if you hover the button on the top left, the surrounding background color changes and reveals that white box:

image

And we don't support dark-mode on SenseStudio yet, so there should be no problem.

corneliusboehm (commented):
The solution for dark mode looks good to me for now :+1: 

image

corneliusboehm (approved):
:lollipop:

PR #115, opened 2021-02-17
[Fix] Make logistic regression safer

YasheshSavani:
Description:

This PR will make running logistic regression on non-tagged videos or on video having only background tags safer, so it will not break the web server.

### DIFF
--- a/tools/sense_studio/sense_studio.py
+++ b/tools/sense_studio/sense_studio.py
@@ -562,9 +562,10 @@ def train_logreg():
 
         X = np.array(X)
         y = np.array(y)
-        logreg = LogisticRegression(C=0.1, class_weight=class_weight)
-        logreg.fit(X, y)
-        dump(logreg, logreg_path)
+        if len(class_weight) > 1:
+            logreg = LogisticRegression(C=0.1, class_weight=class_weight)
+            logreg.fit(X, y)
+            dump(logreg, logreg_path)
 
     return redirect(url_for('annotate', split=split, label=label, path=path, idx=idx))
 


tools/sense_studio/sense_studio.py
```diff
@@ -562,9 +562,10 @@ def train_logreg():
 
         X = np.array(X)
         y = np.array(y)
-        logreg = LogisticRegression(C=0.1, class_weight=class_weight)
-        logreg.fit(X, y)
-        dump(logreg, logreg_path)
+        if len(class_weight) > 1:
```
YasheshSavani, line 565: In `else`, I am thinking of creating a message or something which will be passed to template for the user.

corneliusboehm (commented):
Now thinking about this again, not having a message is probably fine. `len(class_weight) == 1` would mean everything is labeled with the same class, most likely background, so it's fine if we don't predict anything yet.

However I discovered another issue: If you have annotated another class first, the global `logreg` variable will be the `LogisticRegression` object for that. Now if you go to the next class that doesn't have it's own logreg file yet, this part in `show_video_list` will not reset the global variable:
```python
    logreg_path = join(logreg_dir, 'logreg.joblib')
    if os.path.isfile(logreg_path):
        global logreg
        logreg = load(logreg_path)
```
Then in `annotate`, the old `logreg` model will be used for annotating the new class. Could you fix that?

Final thought: The user experience with the "Retrain Model" button is a little weird. You have to submit your current annotations first so they will be used in the model update, but when you hit submit, you will be forwarded to the next video already. I wonder if it would make sense to always retrain the model when you hit submit, or to also include the current annotations in the training :thinking: What do you think?

YasheshSavani:
> Now thinking about this again, not having a message is probably fine. `len(class_weight) == 1` would mean everything is labeled with the same class, most likely background, so it's fine if we don't predict anything yet.
> 
> However I discovered another issue: If you have annotated another class first, the global `logreg` variable will be the `LogisticRegression` object for that. Now if you go to the next class that doesn't have it's own logreg file yet, this part in `show_video_list` will not reset the global variable:
> 
> ```python
>     logreg_path = join(logreg_dir, 'logreg.joblib')
>     if os.path.isfile(logreg_path):
>         global logreg
>         logreg = load(logreg_path)
> ```
> 
> Then in `annotate`, the old `logreg` model will be used for annotating the new class. Could you fix that?
> 
> Final thought: The user experience with the "Retrain Model" button is a little weird. You have to submit your current annotations first so they will be used in the model update, but when you hit submit, you will be forwarded to the next video already. I wonder if it would make sense to always retrain the model when you hit submit, or to also include the current annotations in the training thinking What do you think?

For the `global logreg`, I was facing this situation when I was testing refactoring branch but I fixed that in that branch itself.

corneliusboehm:
> For the global logreg, I was facing this situation when I was testing refactoring branch but I fixed that in that branch itself.

Oh good, then I need to take a look at that again :slightly_smiling_face:

YasheshSavani:
> Now thinking about this again, not having a message is probably fine. `len(class_weight) == 1` would mean everything is labeled with the same class, most likely background, so it's fine if we don't predict anything yet.
> 
> However I discovered another issue: If you have annotated another class first, the global `logreg` variable will be the `LogisticRegression` object for that. Now if you go to the next class that doesn't have it's own logreg file yet, this part in `show_video_list` will not reset the global variable:
> 
> ```python
>     logreg_path = join(logreg_dir, 'logreg.joblib')
>     if os.path.isfile(logreg_path):
>         global logreg
>         logreg = load(logreg_path)
> ```
> 
> Then in `annotate`, the old `logreg` model will be used for annotating the new class. Could you fix that?
> 
> Final thought: The user experience with the "Retrain Model" button is a little weird. You have to submit your current annotations first so they will be used in the model update, but when you hit submit, you will be forwarded to the next video already. I wonder if it would make sense to always retrain the model when you hit submit, or to also include the current annotations in the training thinking What do you think?

Thinking on the second point, will get back to you in sometime on that. :smile:

YasheshSavani:
> Now thinking about this again, not having a message is probably fine. `len(class_weight) == 1` would mean everything is labeled with the same class, most likely background, so it's fine if we don't predict anything yet.
> 
> However I discovered another issue: If you have annotated another class first, the global `logreg` variable will be the `LogisticRegression` object for that. Now if you go to the next class that doesn't have it's own logreg file yet, this part in `show_video_list` will not reset the global variable:
> 
> ```python
>     logreg_path = join(logreg_dir, 'logreg.joblib')
>     if os.path.isfile(logreg_path):
>         global logreg
>         logreg = load(logreg_path)
> ```
> 
> Then in `annotate`, the old `logreg` model will be used for annotating the new class. Could you fix that?
> 
> Final thought: The user experience with the "Retrain Model" button is a little weird. You have to submit your current annotations first so they will be used in the model update, but when you hit submit, you will be forwarded to the next video already. I wonder if it would make sense to always retrain the model when you hit submit, or to also include the current annotations in the training thinking What do you think?

For the second point, I think we should consider `logistic_regression` as an arbitrary or optional thing. Beacause in my view it depends on the users if they want to see tagging time prediction of labels based on the past labelling of videos. If the user only want to do tagging then putting logreg_training as a compulsory task on submit is not appropriate. I don't have a strong opinion on this though. We can discuss with others too.

On the contrary, if we want to do retraining on submit we can do that which will show them predicted labels, but if the users do not want to see that we can have a hide button on the page @manikd31 suggested when I was talking with him. Because if we don't want to see predicted labels we need to delete `logreg.joblib` file everytime instead of doing that having a button click will make it more userfriendly.

corneliusboehm:
Thanks for your input. I have the following suggestion now:
1. We add a switch on the annotation page for showing the current model predictions
2. If it is enabled, the model is always retrained when the "Submit" button is pressed. The "Retrain model" button can be removed.
3. If it is disabled, we hide the prediction labels and we don't retrain the model on "Submit".

We can address this at the end of this sprint if there is time left, otherwise in the next one.

corneliusboehm (approved):
No more errors :+1: UX can be improved in a follow-up.

PR #114, opened 2021-02-16
[Fix] Refactor sense studio.py script

YasheshSavani:
Description:

This PR will modularise the application by splitting `sense_studio.py` into multiple modules/packages like annotations, video recorder etc.

PR #116, opened 2021-02-18
Replace path with project in GET requests

nahuakang:
This PR refactors `project_details` route to use `project` instead of `path`. It should be merged into [PR114](https://github.com/TwentyBN/sense/pull/114) to avoid a more difficult merge conflict resolution.


YasheshSavani (dismissed):
LGTM :+1:

tools/sense_studio/sense_studio.py
```diff
@@ -175,7 +176,7 @@ def project_details(path):
                 'tagged': len(os.listdir(tags_path)) if os.path.exists(tags_path) else 0,
             }
 
-    return render_template('project_details.html', config=config, path=path, stats=stats)
+    return render_template('project_details.html', config=config, project=project, path=path, stats=stats)
```
corneliusboehm, line 179: I think you could also use `config.name` in there to get the project name

tools/sense_studio/annotations.py
```diff
@@ -23,25 +23,27 @@
...
     inference_engine = _load_feature_extractor()
     for split in SPLITS:
         print("\n" + "-" * 10 + f"Preparing videos in the {split}-set" + "-" * 10)
         for label in os.listdir(join(path, f'videos_{split}')):
             compute_frames_features(inference_engine, split, label, path)
-    return redirect(url_for("project_details", path=path))
+    return redirect(url_for("project_details", project=project))
 
 
 @annotations_bp.route('/<split>/<label>/<path:path>/<int:idx>')
```
corneliusboehm, line 49: Apparently there is still a `path` left

nahuakang:
@corneliusboehm @YasheshSavani The `train_logreg` function throws a `ValueError` of `ValueError: This solver needs samples of at least 2 classes in the data, but the data contains only one class: 0` because I have only 1 class. Shall we treat this case more gracefully or?

YasheshSavani:
> @corneliusboehm @YasheshSavani The `train_logreg` function throws a `ValueError` of `ValueError: This solver needs samples of at least 2 classes in the data, but the data contains only one class: 0` because I have only 1 class. Shall we treat this case more gracefully or?

That will be resolved in #115

corneliusboehm (commented):
Working well. The last `path` that is left now is the one in `download_file`. I think if you pass `project` to `frame_annotation.html`, you have all the necessary information in there to reconstruct the path again dynamically. Also then you don't have to lookup the `project` again in `submit_annotation` and `train_logreg`.

corneliusboehm (approved):
:film_projector:

PR #117, opened 2021-02-18
Load existing (saved) annotations for videos

manikd31:
This PR links to [this ticket](https://20bn.atlassian.net/browse/AV-4350) and adds the feature to show the saved annotations for a video, if the user has previously annotated it.


corneliusboehm (approved):
I found a way to integrate this with the regular block of scripts. Working well now :slightly_smiling_face:

PR #118, opened 2021-02-19
Support model selection in tools

corneliusboehm:
## Description
Support using different model versions in all `tools`.

### General Changes
- Extracted features will now be stored per model, so the directory structure looks like this:
```
- features_train/
    - StridedInflatedEfficientNet-pro/
        - num_layers_to_finetune=X/
            - video_0.npy
            - video_1.npy
```
- Methods for obtaining all relevant directories have been added to `tools/utils.py`
- The method `get_weights` was moved into the `ModelConfig` class so that it's also possible to obtain the weights for a specific model config

### SenseStudio
- The best available model is automatically discovered and used for annotation
- In the future we should add a dropdown menu on the UI for selecting from the available models

### Training
- The training script can now be provided with a `model_name` and `model_version`. If those are not provided, the best available model will automatically be used.
- The parameters for a training run will be stored in a `config.json` file. This can later for example be used to look up the correct backbone model.
```py
{
    'backbone_model_name': selected_config.model_name,
    'backbone_model_version': selected_config.version,
    'num_layers_to_finetune': num_layers_to_finetune,
    'classifier': str(gesture_classifier),
    'temporal_training': temporal_training,
    'lr_schedule': lr_schedule,
    'num_epochs': num_epochs,
}
```

### Running custom classifier
- The backbone network that was used for training is now looked up from the `config.json` file

### Conversion
- Now requires `backbone_name` and `backbone_version` for pre-trained classifiers or `path_in` for a custom classifier
- For a custom classifier, the corresponding backbone model will be looked up from the training config


tools/conversion/convert_to_tflite.py
```diff
@@ -127,54 +121,67 @@ def convert(backbone_settings, classifier_settings, output_name, plot_model, ver
     export_keras_to_tflite(keras_file, tflite_file)
 
     if fake_weights:
-        logging.error("************************* Warning!! **************************")
-        logging.error("Weights in checkpoint did not match weights required by network")
-        logging.error("Fake weights were generated where they were needed!!!!!!!!!!!!")
-        logging.error("************************* Warning!! **************************")
+        logging.warning(
```
florianletsch, line 124: This is now a warning again? I think it was an `error` on purpose so that it's not swallowed when the log level is too low.

florianletsch (commented):
Code style changes are looking good! I haven't tested it yet.

tools/conversion/convert_to_tflite.py
corneliusboehm: Oh, we can change it back. I was just confused because the `error` log (which to me is something that is actually breaking) started with the word "warning" (which to me is something that is not breaking)

tools/train_classifier.py
```diff
@@ -113,11 +132,11 @@
         backbone_network.cnn = backbone_network.cnn[0:-num_layers_to_finetune]
 
     # finetune the model
```
sunny-panchal, line 134:
```suggestion
    # Fine-tune the model
```

tools/train_classifier.py
```diff
@@ -130,17 +149,19 @@
     extractor_stride = backbone_network.num_required_frames_per_layer_padding[0]
 
     # create the data loaders
```
sunny-panchal, line 151:
```suggestion
    # Create the data loaders
```

tools/train_classifier.py
```diff
@@ -130,17 +149,19 @@
...
-    valid_loader = generate_data_loader(path_in, f"features_valid_num_layers_to_finetune={num_layers_to_finetune}",
-                                        "tags_valid", label_names, label2int, label2int_temporal_annotation,
+    features_dir = utils.get_features_dir(path_in, 'valid', selected_config, num_layers_to_finetune)
+    tags_dir = utils.get_tags_dir(path_in, 'valid')
+    valid_loader = generate_data_loader(features_dir, tags_dir, label_names, label2int, label2int_temporal_annotation,
                                         num_timesteps=None, batch_size=1, shuffle=False, stride=extractor_stride,
                                         temporal_annotation_only=temporal_training)
 
-    # modeify the network to generate the training network on top of the features
+    # modify the network to generate the training network on top of the features
```
sunny-panchal, line 164:
```suggestion
    # Modify the network to generate the training network on top of the features
```

tools/train_classifier.py
```diff
@@ -167,6 +188,24 @@
 
     lr_schedule = {0: 0.0001, 40: 0.00001}
     num_epochs = 80
+
+    # Save training config and label2int dictionary
+    config = {
```
sunny-panchal, line 193: Should we save the date/time the model was trained too?

sunny-panchal (commented):
I haven't tested it yet but looks good on first pass

tools/train_classifier.py
corneliusboehm: Sure, I'll add a start and end time

tools/train_classifier.py
corneliusboehm: I also prefer this spelling of "fine-tune", but so far we have written "finetune" everywhere else in the project :shrug:

sense/loading.py
```diff
@@ -53,9 +53,27 @@ def __init__(self, model_name: str, version: str, feature_converters: List[str])
         self.version = version
         self.feature_converters = feature_converters
 
-    def get_path_weights(self):
+    @property
+    def combined_model_name(self):
```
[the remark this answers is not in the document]
corneliusboehm: Probably not, I could make it a regular attribute

sense/loading.py
```diff
@@ -89,26 +107,33 @@ def get_relevant_weights(model_config_list: List[ModelConfig], requested_model_n
...
+def load_backbone_model_from_config(checkpoint_path: str) -> Tuple[ModelConfig, dict]:
+    """
+    Load the backbone model that was used in training for the given model checkpoint as indicated in the 'config.json'
+    file. If there is no config file, StridedInflatedEfficientNet-pro will be used per default.
+    """
+    config_file = os.path.join(checkpoint_path, 'config.json')
+    if os.path.exists(config_file):
+        with open(config_file, 'r') as cf:
+            config = json.load(cf)
+            backbone_model_config = ModelConfig(config['backbone_model_name'], config['backbone_model_version'], [])
```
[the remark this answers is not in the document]
corneliusboehm: In our example scripts, the command line parameters are called `--model_name` and `--model_version`. Should we change those to `--backbone_name` as well?
cc @guillaumebrg

README.md
```diff
@@ -275,13 +275,13 @@ models to the TensorFlow Lite format.
 Our models can be converted to TensorFlow Lite using the following script:
 
 ```shell
-python tools/conversion/convert_to_tflite.py --backbone=StridedInflatedEfficientNet --classifier=efficient_net_gesture_recognition --output_name=model
+python tools/conversion/convert_to_tflite.py --backbone_name=StridedInflatedEfficientNet --backbone_version=pro --classifier=gesture_recognition --output_name=model
```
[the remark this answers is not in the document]
corneliusboehm: If a custom trained model is used, the backbone model and version are inferred from the training config. If one of our pre-trained models is converted, we somehow need to provide that information though.

@guillaumebrg What do you think about merging the version into the name or calling it variant instead?

tools/conversion/convert_to_tflite.py
corneliusboehm: As discussed, we should make sure that the default log level includes warnings

tools/sense_studio/sense_studio.py
```diff
@@ -19,7 +19,8 @@
 from flask import request
 from flask import url_for
 
-from tools.sense_studio import utils
+from tools import utils
+from tools.sense_studio import utils as studio_utils
```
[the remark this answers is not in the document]
corneliusboehm: I have at least renamed one of the `utils` modules now, so we no longer have that conflict :grin:

corneliusboehm:
Ready for a final review, so we can merge what we have so far.

These questions are still open and can be addressed later:
- Call `--backbone_version` `--backbone_variant` instead to avoid confusion with a numerical version? Or merge version into `--backbone_name`, so that users only have to indicate one?
- In the example scripts, the command line parameters are called `--model_name` and `--model_version`, while the tools call it `--backbone_name`. Should we unify those?
- Should we shorten all model names to  `SIEN/SI-EN` and `SIMN/SI-MN`?
- "Fine-tune" instead of "finetune"?

guillaumebrg:
I haven't finished reviewing yet... but regarding your questions:

> Call --backbone_version --backbone_variant instead to avoid confusion with a numerical version? Or merge version into --backbone_name, so that users only have to indicate one?

No strong opinion here. I'm fine with all 3 solutions (leave it like, rename to "backbone_variant" or merge into "backbone_name"). I suggest we leave it like this in this PR and change that in another PR.

> In the example scripts, the command line parameters are called --model_name and --model_version, while the tools call it --backbone_name. Should we unify those?

Yes, I think that would be good.

> Should we shorten all model names to SIEN/SI-EN and SIMN/SI-MN?

I vote "yes".

> "Fine-tune" instead of "finetune"?

No strong opinion

PR #119, opened 2021-02-19
[Feature] create navigation buttons on annotations page

YasheshSavani:
Description:

This PR will create two new buttons Next and Previous on the frame annotations page for users to move around videos without doing annotations.


corneliusboehm (dismissed):
Working well :+1:

corneliusboehm (approved):

PR #120, opened 2021-02-22
Load classifier checkpoint when resuming training

corneliusboehm:
When resuming training from a checkpoint, we were previously only using the trained feature extractor weights, but not the ones for the classifier. This is fixed now.

Furthermore, I added a flag for always overwriting existing files (`--overwrite`) and simplified the code for that check a little bit.


sunny-panchal (approved):
MSTM :+1:

yoga-0125 (approved):
LGTM

PR #121, opened 2021-02-23
[Feature] Navigation bar with different modules

YasheshSavani:
Description:

This PR will create a navigation bar to access different functionality of project on every page.


tools/sense_studio/annotation.py
```diff
@@ -52,7 +54,7 @@ def show_video_list(project, split, label):
 
     video_list = zip(videos, tagged, list(range(len(videos))))
     return render_template('video_list.html', video_list=video_list, split=split, label=label, path=path,
-                           project=project)
+                           project=project, labels=labels)
```
YasheshSavani, line 57: Thinking of a better name then `labels`.

tools/sense_studio/templates/navigation.html
```diff
@@ -0,0 +1,66 @@
+{% block navigation %}
+    <a href="{{ url_for('project_details', project=project) }}" class="item">Setup Class</a>
```
manikd31, line 2: Can we have something like `Project Overview` instead of `Setup Class`? The latter is a bit unclear. Also, this may be addressed when we have a decent UI design to work with, so it's not a priority.

corneliusboehm:
Haven't looked at the code yet, but here's some feedback on the UI:
- The buttons for Training, Evaluation and Conversion should be disabled and maybe show a tooltip that they will soon be added
- We need to find a uniform naming scheme for all those buttons, because currently it's a mix of verbs and nouns. My suggestion would be `Projects -> Project Details -> Video Recording -> Annotation -> Training -> Testing -> Conversion`
- Projects could also be a dropdown that allows switching to the other projects :thinking:

corneliusboehm:
I did a bit of research on how to pass some information to all templates. It looks like you can use `@app.context_processor` for that or maybe store the data in `flask.session` or `flask.g`: https://flask.palletsprojects.com/en/1.1.x/templating/

YasheshSavani:
> Haven't looked at the code yet, but here's some feedback on the UI:
> 
>     * The buttons for Training, Evaluation and Conversion should be disabled and maybe show a tooltip that they will soon be added. Best would be to even hint at the scripts that you have to run for now.
> 
>     * We need to find a uniform naming scheme for all those buttons, because currently it's a mix of verbs and nouns. My suggestion would be `Projects -> Project Details -> Video Recording -> Annotation -> Training -> Testing -> Conversion`
> 
>     * Projects could also be a dropdown that allows switching to the other projects thinking

we can do project switching, but I think Sunny and Antoine are working on the design so let's wait for that?

sunny-panchal:
> > Haven't looked at the code yet, but here's some feedback on the UI:
> > ```
> > * The buttons for Training, Evaluation and Conversion should be disabled and maybe show a tooltip that they will soon be added. Best would be to even hint at the scripts that you have to run for now.
> > 
> > * We need to find a uniform naming scheme for all those buttons, because currently it's a mix of verbs and nouns. My suggestion would be `Projects -> Project Details -> Video Recording -> Annotation -> Training -> Testing -> Conversion`
> > 
> > * Projects could also be a dropdown that allows switching to the other projects thinking
> > ```
> 
> we can do project switching, but I think Sunny and Antoine are working on the design so let's wait for that?

The design will probably be revised after the next hackathon so for now we can leave the project switching to happen from the page vs as a dropdown.

tools/sense_studio/templates/navigation.html
```diff
@@ -52,13 +52,13 @@
         </div>
     </div>
 </div>
-<a href="#" class="item">Train Model</a>
-<a href="#" class="item">Evaluate Model</a>
+<a href="javascript:void(0)" class="item hashoverpopup" data-content="Coming Soon! Hint: Use train_classifier.py" data-position="bottom left" data-variation="basic">Training</a>
```
YasheshSavani, line 55: Tell me if `data-content` need to be more informative, right now the content is basic hint.

tools/sense_studio/annotation.py
```diff
@@ -53,7 +55,7 @@ def show_video_list(project, split, label):
...
```
corneliusboehm, line 58: Is this still necessary now that we have a context processor?

tools/sense_studio/templates/navigation.html
```diff
@@ -0,0 +1,70 @@
...
+    Video Recording
+    <i class="dropdown icon"></i>
+    <div class="menu">
+        <div class="item">
+            <i class="dropdown icon"></i>
+            Train
+            <div class="menu">
+                {% for label in class_labels  %}
+                    <a href="{{ url_for('video_recording_bp.record_video', project=project, split='train',
+                     label=label) }}" class="item">{{label}}</a>
```
corneliusboehm, line 13: Weird line break. I don't think we have a character limit on html files yet.
And please add spaces around the variable: `{{  label  }}`

tools/sense_studio/sense_studio.py
```diff
@@ -268,5 +268,14 @@ def add_header(r):
     return r
 
 
+@app.context_processor
+def class_labels_processor():
+    def inject_class_labels(project):
+        path = utils.lookup_project_path(project)
+        labels = utils.get_class_labels(path)
+        return labels
+    return dict(inject_class_labels=inject_class_labels)
```
corneliusboehm, line 277: Very interesting :thinking: Reading the example on https://flask.palletsprojects.com/en/1.1.x/templating/#context-processors, we might also have used some global scope to store the current `project` name, but your solution also works. Maybe just add a comment on how this works and is used, so others can understand more easily.

tools/sense_studio/templates/navigation.html
```diff
@@ -0,0 +1,70 @@
...
+            <i class="dropdown icon"></i>
+            Valid
+            <div class="menu">
+                {% for label in class_labels %}
+                    <a href="{{ url_for('video_recording_bp.record_video', project=project, split='valid',
+                     label=label) }}" class="item">{{label}}</a>
+                {% endfor %}
+            </div>
+        </div>
+
```
corneliusboehm, line 27: Weird empty line

corneliusboehm (commented):
Looking good and the solution with the context processor works well :+1: Just a few minor comments left.

tools/sense_studio/templates/video_list.html
```diff
@@ -1,7 +1,11 @@
-{% extends 'skeleton.html' %}
+{% extends 'project_details.html' %}
```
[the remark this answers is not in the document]
YasheshSavani: Sorry, I forgot to update that file at the end

tools/sense_studio/annotation.py
YasheshSavani: Not need, same for this as video_list.html

tools/sense_studio/sense_studio.py
YasheshSavani: sure

tools/sense_studio/templates/project_details.html
```diff
@@ -17,7 +21,7 @@ <h1 class="ui header">
         <i class="file video outline icon"></i>
         Prepare Annotations
     </a>
-    <input type="hidden" id="projectName" value="{{ config.name }}">
+    <input type="hidden" id="projectName" value="{{ project }}">
```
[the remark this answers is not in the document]
YasheshSavani: I'll do that thanks for pointing it out.

tools/sense_studio/sense_studio.py
```diff
@@ -268,12 +268,15 @@ def add_header(r):
     return r
 
 
+# This context processor will inject a method into templates,
+# Which can be invoked like an ordinary method in HTML templates.
+# E.g. check navigation.html: line 1-2
```
corneliusboehm, line 273: This is exactly what a docstring should be used for, i.e. triple quotes below the function signature

corneliusboehm (approved):

PR #123, opened 2021-02-25
Add unit test for sense/camera.py

yoga-0125:
create tests/test_camera.py for sense/camera.py unit test.


tests/test_camera.py
```diff
@@ -0,0 +1,137 @@
+import unittest
```
sunny-panchal, line 1: Should import in this order (with blank lines after each group and alphabetical ordering within groups): 
```
1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.
```
 https://www.python.org/dev/peps/pep-0008/#imports

tests/test_camera.py
```diff
@@ -0,0 +1,137 @@
...
+import sense.camera as camera
+import os
+import cv2
+
+
+class TestGetImage(unittest.TestCase):
+
+    def setUp(self) -> None:
+        cwd = os.getcwd()
+        video_path = os.path.join(cwd, 'tests/resources/test_video.mp4')
```
sunny-panchal, line 12: The paths are used in multiple places and can be set once after the imports as a global variable.

tests/test_camera.py
```diff
@@ -0,0 +1,137 @@
+import unittest
+from unittest.mock import patch
+import sense.camera as camera
+import os
+import cv2
+
+
+class TestGetImage(unittest.TestCase):
```
sunny-panchal, line 8: You can rename this class to `TestVideoSource` and include all the methods in the other classes that are testing `VideoSource` -- all of them use the same `setup()` function. 

The same could then be done with a class for `TestVideoStream` and `TestVideoWriter`

sunny-panchal (commented):
Some initial comments:

tests/test_camera.py
yoga-0125: Thanks, I've updated them.

tests/test_camera.py
yoga-0125: The paths have updated as global variables.

tests/test_camera.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class TestVideoSource(unittest.TestCase):
+
+    def setUp(self) -> None:
+        self.video = camera.VideoSource(filename=video_path)
+
+    @patch('sense.camera.VideoSource.pad_to_square')
+    def test_get_image(self, mock_pad_to_square):
+        square_img_path = os.path.join(cwd, 'tests/resources/square_SENSE.png')
```
corneliusboehm, line 20: Paths should be system-independent if possible and `os.path.join` is already the right tool for that:
```python
square_img_path = os.path.join(cwd, 'tests', 'resources', 'square_SENSE.png')
```

tests/test_camera.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_stop(self):
+        self.stream.stop()
+        self.assertTrue(self.stream._shutdown)
+
+    def test_extract_image(self):
+        img_tuple = self.video.get_image()
+        self.stream.frames.put(img_tuple, False)
+        imgs = self.stream.get_image()
+        assert str(type(imgs[0])) == "<class 'numpy.ndarray'>"
```
corneliusboehm, line 53: Why not directly check for the correct class?
```python
assert type(imgs[0]) == np.ndarray
```

yoga-0125: > Why not directly check for the correct class?
> 
> ```python
> assert type(imgs[0]) == np.ndarray
> ```

update

tests/test_camera.py
```diff
@@ -1,13 +1,13 @@
-import os
 import unittest
 from unittest.mock import patch
 
 import cv2
+import numpy as np
+from pathlib import Path
```
corneliusboehm, line 6: Oh, `pathlib` again. That's still an ongoing discussion, but I'm all for it :grin:

yoga-0125: > pathlib

Thanks for the explanation. Which module is reliable (stable) to use (os.path or pathlib or others)?

corneliusboehm: Both are stable and reliable, but there are other pros and cons to them. It's just an internal discussion and we haven't completely decided yet.

sunny-panchal: Let's stick to `os.path` for now to keep it in line with the rest of the project

tests/test_camera.py
```diff
@@ -0,0 +1,89 @@
+import unittest
+from unittest.mock import patch
+
+import cv2
+import numpy as np
+from pathlib import Path
+
+import sense.camera as camera
+
+video_path = str(Path('tests/resources/test_video.mp4').resolve())
```
sunny-panchal, line 10:
```suggestion
VIDEO_PATH = str(Path('tests/resources/test_video.mp4').resolve())
```
variable name should be all caps when its a global constant

tests/test_camera.py
yoga-0125: Ok

tests/test_camera.py
```diff
@@ -53,3 +57,56 @@ def read(self):
...
+
+
+class TestVideoWriter(unittest.TestCase):
+
+    def setUp(self) -> None:
+        self.output_video_path = os.path.join(os.path.dirname(__file__), 'resources', 'test_writer.mp4')
+        self.videowriter = VideoWriter(path=self.output_video_path, fps=12.0, resolution=(40, 30))
+
+    def test_write(self):
+        input_video_path = VIDEO_PATH
```
sunny-panchal, line 91: This variable is only used once in the line below so rather than assigning it here you can just pass the path in directly

tests/test_camera.py
```diff
@@ -53,3 +57,56 @@ def read(self):
...
+        input_video = cv2.VideoCapture(input_video_path)
+        output_video_path = self.output_video_path
```
sunny-panchal, line 93: Same with this variable. Only using it once below so you could just assign the attribute there directly.

tests/test_camera.py
```diff
@@ -53,3 +57,56 @@ def read(self):
...
+
+        while True:
+            ret, frame = input_video.read()
+            self.videowriter.write(frame)
+            if not ret:
+                break
+
+        input_video.release()
+        self.videowriter.release()
+        output_video = cv2.VideoCapture(output_video_path)
```
sunny-panchal, line 103: To test whether the file is written, you open it with `VideoCapture` then assert that it indeed did. Instead, since `test_video.mp4` is the one being written again without modification, would it make sense to check that both files are the same and something hasn't modified/corrupted it in the process?

sunny-panchal (commented):
Left a few comments; some small changes requested otherwise looks good :+1: 

The files `SENSE.png` and `square_SENSE.png` are unused and can be removed.

tests/test_camera.py
yoga-0125: Thanks, I've removed `input_video_path`

tests/test_camera.py
yoga-0125: Which variable?

sunny-panchal: `output_video_path`

sunny-panchal (approved):
looks good, thanks for the changes :+1:

tests/test_camera.py
```diff
@@ -53,3 +57,61 @@ def read(self):
...
+        self.videowriter.release()
+        output_video = cv2.VideoCapture(self.output_video_path)
+
+        # Check whether new output file is the same as the original
+        while True:
+            ret, frame_in = input_video.read()
+            _, frame_out = output_video.read()
+            if not ret:
+                break
+            assert np.all([frame_in, frame_out])
```
corneliusboehm, line 109: This is not an equality test. It just tests that both arrays are not `None`.

tests/test_camera.py
```diff
@@ -53,3 +57,61 @@ def read(self):
...
+                break
+
+        input_video.release()
+        self.videowriter.release()
+        output_video = cv2.VideoCapture(self.output_video_path)
+
+        # Check whether new output file is the same as the original
+        while True:
+            ret, frame_in = input_video.read()
+            _, frame_out = output_video.read()
```
corneliusboehm, line 106: As you have committed the output video to the resources folder, this test case would also work without any new output video being written. It would be cleaner to not commit the video and always remove it after it has been written in this test case.

tests/test_camera.py
sunny-panchal: Whoops, sorry, that was my mistake--fixing now

PR #124, opened 2021-02-25
[Feature] GPU option for sense_studio.py

manikd31:
This PR adds a toggle switch to the top of `project_details` page where the user can change using/not using the GPU. The `use_gpu` parameter will be saved inside the `project_config.json` and will be reloaded for every new session with the respective project.

---

**_EDIT:_** Removing the global inference_engine object from inside `utils.py` will make the GPU option more dynamic, that is, it'll overcome the one-time-only switch for preparing annotations.

---

**Not Using GPU:**

Not Using GPU

---

**Using GPU:**

Using GPU


corneliusboehm (approved):
I've made a few changes like not requiring a page refresh when the GPU status is toggled and storing the GPU status permanently in the project config. We can talk those through when you're online, but from my side this is good to go now :+1:

corneliusboehm (approved):
LGTM

PR #129, opened 2021-03-04
[Feature] Option to use "Assisted Tagging" feature

manikd31:
**Ticket:** [AV-4412](https://20bn.atlassian.net/browse/AV-4412)

This PR adds another option to the existing project settings which can be toggled to show/hide the logistic regression model predictions on the frame-annotation page. Enabling/disabling the toggle for "Assisted Tagging" will reload the page and show/hide the most up-to-date Logistic Regression predictions.


corneliusboehm (commented):
Looking good and working :+1: 

I observed only one issue: When I had annotated a video with the option disabled and then enabled it, the predictions did not show up. I had to re-submit the same annotations to see something. Would it make sense to retrain the logreg if the option is enabled and you are already on that page or enter it again?

And we talked about the fact before, that features only need to be computed if that option is enabled. Is that still planned?

manikd31:
Thanks for the update. I'll be working on this further on, and yes, it's still planned to compute features only if the option is enabled.

tools/sense_studio/annotation.py
```diff
@@ -87,7 +84,8 @@ def prepare_annotation(project):
             compute_frames_features(inference_engine=inference_engine,
                                     videos_dir=videos_dir,
                                     frames_dir=frames_dir,
-                                    features_dir=features_dir)
+                                    features_dir=features_dir,
+                                    with_features=False)
```
corneliusboehm, line 88: Shouldn't `with_features` depend on the setting?

tools/sense_studio/annotation.py
```diff
@@ -112,20 +110,19 @@ def annotate(project, split, label, idx):
...
+              if utils.is_image_file(image)]
 
     # Load logistic regression model if available
     logreg_path = os.path.join(logreg_dir, 'logreg.joblib')
-    if os.path.isfile(logreg_path):
+    features_path = os.path.join(features_dir, f'{videos[idx]}.npy')
+    if os.path.isfile(logreg_path) and os.path.isfile(features_path):
         logreg = load(logreg_path)
+        features = np.load(os.path.join(features_dir, f'{videos[idx]}.npy')).mean(axis=(2, 3))
         classes = list(logreg.predict(features))
```
corneliusboehm, line 123: Shouldn't this part only be performed, if the setting is enabled?

tools/sense_studio/static/main.js
```diff
@@ -253,6 +253,21 @@ function toggleMakeProjectTemporal(path) {
...
 
 
+function toggleShowPredictions(path, split, label) {
+    let logregInput = document.getElementById('logregInput');
+    response = syncRequest('/toggle-project-setting', {path: path, setting: 'show_logreg', split: split, label: label});
+
+    // Show/hide all LogReg prediction-labels
+    if (response.setting_status) {
+        logregInput.setAttribute('checked', 'checked');
+        $('.logreg-predictions').show();
```
corneliusboehm, line 263: The logreg labels haven't been updated yet, so they are not displaying the outputs of the new logreg model that has just been trained

tools/sense_studio/templates/frame_annotation.html
```diff
@@ -8,6 +8,16 @@
...
 {% block main %}
 <div id="main" class="ui main container">
+    <div class="ui segment right floated">
+        <h3 class="ui header center aligned">Settings</h3>
+        <div class="ui toggle checkbox hashoverpopup"
+           data-content="Train a simple logistic regression model live on temporal annotations and show current predictions?"
+           data-position="left center">
+            <input type="checkbox" id="logregInput" {% if project_config.show_logreg %} checked="checked" {% endif %}
+               onclick="toggleShowPredictions('{{ path }}', '{{ split }}', '{{ label }}');">
+            <label>Show Predictions</label>
```
corneliusboehm, line 18: Guillaume recently called this "Assisted tagging" and I really like that term, because it highlights the purpose of this feature more. Can we use that name throughout the code?

corneliusboehm (commented):
I think there is only one missing detail: When the user enables the switch on the annotation page, the logreg model is retrained, but the labels on the page are not updated yet, so either you're displaying old ones or none If there were no predictions yet (can be tested by creating a new class).

By the way, you seem to accidentally have committed a `reactive_gesture_test_1.mp4`

sense/finetuning.py
```diff
@@ -234,6 +236,8 @@ def compute_frames_features(inference_engine: InferenceEngine, videos_dir: str,
         numbered .jpg files in there.
     :param features_dir:
         Directory where computed features should be stored. One .npy file will be created per video.
+    :param with_features:
+        Whether to extract features from frames using logreg or not.
```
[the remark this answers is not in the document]
manikd31: Removed the `with_features` parameter, and now it's based on the project-setting status.

sense/finetuning.py
```diff
@@ -188,23 +188,24 @@ def compute_features(video_path, path_out, inference_engine, num_timesteps=1, pa
...
-    predictions = np.concatenate([temporal_dependency_features, predictions], axis=0)
-    features = np.array(predictions)
-    os.makedirs(os.path.dirname(path_out), exist_ok=True)
-    np.save(path_out, features)
+        # predictions of the actual video frames
+        predictions = inference_engine.infer(clip[:, frames_to_add + 1:], batch_size=batch_size)
+        predictions = np.concatenate([temporal_dependency_features, predictions], axis=0)
+        features = np.array(predictions)
+        os.makedirs(os.path.dirname(path_out), exist_ok=True)
+        np.save(path_out, features)
```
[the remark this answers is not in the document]
manikd31: Now, we have two separate methods `compute_only_frames` and `compute_only_features`, both of which are called by the parent method `compute_frames_features`.

tools/sense_studio/annotation.py
manikd31: Updated this to use the project-setting

tools/sense_studio/static/main.js
manikd31: The new method reloads the page to display the newly trained LogReg predictions.

tools/sense_studio/annotation.py
manikd31: Done.

manikd31:
> By the way, you seem to accidentally have committed a `reactive_gesture_test_1.mp4`

Apologies for that, it has now been removed from the branch.

corneliusboehm (approved):
Tested - everything seems to be working well :heavy_check_mark:

PR #132, opened 2021-03-10
[Feature] Training Module

YasheshSavani:
Description:

This PR will add a feature in Sense Studio to train your model from the web-server where user can do model selection, enter number of epochs, enter number of layers to finetune, and define where to save the checkpoint.


tools/sense_studio/static/main.css
```diff
@@ -70,3 +70,14 @@ a {
...
 }
+
+.ui.segment.train-console {
+    overflow: auto;
+    height: 500px;
+    background-color: black;
+    color: white;
+}
+.font-style {
+    font-family: Ubuntu Mono;
```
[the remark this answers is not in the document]
YasheshSavani: Thank you, I'll keep in mind :+1:

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+def train_model():
+    data = request.form
+    project = data['project']
+    num_layers_to_finetune = data['layers_to_finetune']
+    path_out = data['output_folder']
+    model_name = data['model_name']
+    path = utils.lookup_project_path(project)
+    config = utils.load_project_config(path)
+
+    train_classifier = ["python tools/train_classifier.py", f"--path_in=/home/twentybn/Code/sense/dataset/{project}",
```
corneliusboehm, line 45: :warning: That's why it doesn't work for me! You have hard-coded your own dataset path.

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+                        f"--num_layers_to_finetune={num_layers_to_finetune}",
+                        "--use_gpu" if config['use_gpu'] else "",
+                        f"--path_out={path_out}" if path_out else "",
```
corneliusboehm, line 48: I was somehow expecting the output folder to be appended to `<project_path>/checkpoints`.
We could even indicate that on the input field by using the labeled one with the path prefix: https://semantic-ui.com/elements/input.html#labeled

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+                        f"--model_name={model_name}" if model_name else "",
```
corneliusboehm, line 49: We need to indicate `model_name` and `model_version`

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+                        "--overwrite"]
```
corneliusboehm, line 50: If the project is temporal, we need to include the `--temporal_training` flag and the `--path_annotations` parameters

corneliusboehm, line 50: You are not passing the number of `epochs` to the script

tools/sense_studio/train_module.py
corneliusboehm, line 45: The path to the `train_classifier` script should also be indicated system-independent, i.e. using `os.path.join`

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+
+    train_classifier = ' '.join(train_classifier)
+    train_classifier = shlex.split(train_classifier)
```
corneliusboehm, line 53: Why exactly is `shlex` needed here?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+
+    global PROCESS
+    PROCESS = subprocess.Popen(train_classifier, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+
+    def generate():
+        while True:
+            output = PROCESS.stdout.readline()
+            if output == b'' and PROCESS.poll() is not None:
```
corneliusboehm, line 61: Can you explain that end condition?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+                        f"--model_name={model_name}" if model_name else "",
+                        "--overwrite"]
+
+    train_classifier = ' '.join(train_classifier)
+    train_classifier = shlex.split(train_classifier)
+
+    global PROCESS
+    PROCESS = subprocess.Popen(train_classifier, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+
+    def generate():
```
corneliusboehm, line 58: Please use a specific name for this function

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+            output = PROCESS.stdout.readline()
+            if output == b'' and PROCESS.poll() is not None:
+                PROCESS.terminate()
+                break
+            if output:
+                time.sleep(0.1)
+                yield output.decode().strip() + '\n'
+
+    return flask.Response(stream_with_context(stream_template('train.html', project=data['project'], is_disabled=True,
+                                                              models=utils.BACKBONE_MODELS, logs=generate())))
```
corneliusboehm, line 69: Is it somehow possible to do this without a page refresh? Currently, all the input fields are reset when the training starts, so you don't see the parameters that you entered, and the URL also changes, even though it looks like the same page.

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+
+    global PROCESS
+    if PROCESS:
+        PROCESS.terminate()
+        PROCESS = None
+        log = "Training Cancelled."
+    else:
+        log = "No Training Process Running to Terminate."
+    return render_template('train.html', project=data['project'], models=utils.BACKBONE_MODELS, is_disabled=False,
+                           logs=[log])
```
corneliusboehm, line 84: If we use a `redirect`, the URL at least doesn't show `training/cancel-training`

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+    train_classifier = shlex.split(train_classifier)
+
+    global PROCESS
+    PROCESS = subprocess.Popen(train_classifier, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+
+    def generate():
+        while True:
+            output = PROCESS.stdout.readline()
+            if output == b'' and PROCESS.poll() is not None:
+                PROCESS.terminate()
```
corneliusboehm, line 62: Shouldn't we also set `PROCESS = None` in this case?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+                        "--use_gpu" if config['use_gpu'] else "",
+                        f"--path_out={path_out}" if path_out else "",
+                        f"--model_name={model_name}" if model_name else "",
+                        "--overwrite"]
+
+    train_classifier = ' '.join(train_classifier)
+    train_classifier = shlex.split(train_classifier)
+
+    global PROCESS
+    PROCESS = subprocess.Popen(train_classifier, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
```
corneliusboehm, line 56: What happens if the user leaves the page, before training has ended and without clicking `Cancel`? Will there just be an invisible long-running process in the backend?

Can we maybe indicate on the UI that there is currently a training in progress so the user can go there to cancel it?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+from flask import render_template
+from flask import request
+from flask import current_app
+from flask import stream_with_context
+
+from tools.sense_studio import utils
+
+train_bp = Blueprint('train_bp', __name__)
+
+PROCESS = None
```
corneliusboehm, line 18: How about `training_process`?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+
+    global PROCESS
+    PROCESS = subprocess.Popen(train_classifier, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False)
+
+    def generate():
+        while True:
+            output = PROCESS.stdout.readline()
+            if output == b'' and PROCESS.poll() is not None:
+                PROCESS.terminate()
+                break
```
corneliusboehm, line 63: When training is done, the `Train Model` button is still disabled

tools/sense_studio/templates/train.html
```diff
@@ -0,0 +1,63 @@
+{% extends 'skeleton.html' %}
```
corneliusboehm, line 1: I would suggest `training.html`

tools/sense_studio/templates/train.html
```diff
@@ -0,0 +1,63 @@
...
+      <button class="ui button" type="submit" tabindex="1" formaction="{{ url_for('.cancel_training') }}">Cancel</button>
+    </form>
+    <div class="ui segment train-console">
+        {% if is_disabled %}
+            <h4 class="font-style">Training Model...</h4>
+        {% endif %}
+        <h4 class="font-style">
+            {% for log in logs %}
+                {{ log }} <br />
+            {% endfor %}
```
corneliusboehm, line 59: Is it possible to make this area auto-scroll to the most recent line?

tools/sense_studio/train_module.py
```diff
@@ -0,0 +1,84 @@
...
+    project = urllib.parse.unquote(project)
+    return render_template('train.html', project=project, models=utils.BACKBONE_MODELS, is_disabled=False)
+
+
+def stream_template(template_name, **context):
+    # Ref: https://flask.palletsprojects.com/en/1.1.x/patterns/streaming/
+    current_app.update_template_context(context)
+    t = current_app.jinja_env.get_template(template_name)
+    rv = t.stream(context)
+    return rv
```
corneliusboehm, line 32: It looks like because of this streaming mechanism, the footer of the page is only rendered when the training is done. I'm not sure yet, if this streaming functionality is the best solution for our case, because it is originally intended for splitting up an "enormous amount of data", not for updating delayed information.

tools/sense_studio/static/main.js
```diff
@@ -251,3 +253,17 @@ function toggleMakeProjectTemporal(path) {
         $('.temporal').hide();
     }
 }
+
+// TODO: Not Working Properly (Needs investigation)
+function disableTrainModel() {
+    let trainModel = document.getElementById('trainModel');
+    trainModel.setAttribute('disabled', 'disabled');
+    $('.train-console').show();
+}
```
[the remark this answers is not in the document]
YasheshSavani: did you see my new approach for this of sending disabled from backend as the behaviour will be fixed?

corneliusboehm:
Lastly, I think I would still like to extract the whole training script into a python function that we can then execute via `threading` or `multiprocessing`. This way, we're safer against changes to that script and don't rely on building a big sequence of command line arguments.

tools/sense_studio/templates/train.html
YasheshSavani: Finding a way on that already have in mind

tools/sense_studio/train_module.py
YasheshSavani: Have an idea on that using the new approach of sending `disabled` from backend

tools/sense_studio/templates/train.html
```diff
@@ -0,0 +1,63 @@
...
+                </div>
+            <div class="field train">
+                <label>Number of layers to finetune</label>
+                <input type="number" name="layers_to_finetune" min="5" value="5">
+            </div>
+        </div>
+        <div class="two fields wide">
+            <div class="field train">
+                <label>Epochs</label>
+                <input type="number" name="epochs" min="80" value="80">
```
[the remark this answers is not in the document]
YasheshSavani: The default cases are still TBD and to be updated in the ticket. Right now, it is based on default values given in the `train_classifier` script itself.

tools/sense_studio/train_module.py
YasheshSavani: I forgot to remove the before pushing it up, sorry.

tools/sense_studio/train_module.py
YasheshSavani: Here, the readline() will give you an encoded string in return, and `PROCESS.poll()` returns `None` if the process is still running, if the process is completed then it will return anything like `-1` or `0`. 
So here this `and` between them helps if there is an empty line in the console log but the processing is not complete yet it will still going on.

tools/sense_studio/train_module.py
YasheshSavani: I'll investigate on that but afaik if the user leaves that page the training process automatically gets cancelled. I have felt that but still will investigate.

tools/sense_studio/train_module.py
YasheshSavani: I'll investigate more on `multiprocessing` and `threading` if we can do that

corneliusboehm: Here is an example using `EventSource`, which I think should work well for us: https://stackoverflow.com/a/13388915/6611214

tools/sense_studio/static/main.js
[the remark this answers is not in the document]
corneliusboehm: Yes, so can this code be removed?

tools/sense_studio/train_module.py
corneliusboehm: So wouldn't it be enough to only check `PROCESS.poll() is not None`, or is there a case where `poll` returns `None` but `readline` returned a non-empty string?

YasheshSavani: Yes, there is a case. It will happen when processing is done but the console logs are being saved in `stdout` buffer from which we are reading line one by one for display. So, here processing and reading from buffer are independent processes, so it can happen that process might be complete but we still have some lines left to read from stdout.

tools/sense_studio/static/main.js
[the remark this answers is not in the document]
YasheshSavani: Yes

tools/sense_studio/train_module.py
YasheshSavani: I referred the same link and you can see that after this `EventSource` approach the user  has  also posted `stream_template` approach.

corneliusboehm: Yes, but is there an advantage to the `stream_template` approach?

YasheshSavani: I have not tried `EventSource` in the first place and `stream_template` is provided by Flask itself. That's why I thought of using it as flask can handle its own thing properly. But there is no problem in investigating new thing which makes the feature as we want.

tools/sense_studio/templates/train.html
```diff
@@ -0,0 +1,63 @@
...
+                            {% endfor %}
+                        </div>
+                    {% else %}
+                       <div class="default text">No models to select</div>
+                    {% endif %}
+                    </div>
+                </div>
+            <div class="field train">
+                <label>Number of layers to finetune</label>
+                <input type="number" name="layers_to_finetune" min="5" value="5">
```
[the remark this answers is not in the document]
YasheshSavani: Max is still TBD

tools/sense_studio/utils.py
```diff
@@ -18,6 +19,8 @@
     ModelConfig('StridedInflatedMobileNetV2', 'lite', []),
 ]
 
+BACKBONE_MODELS = list(MODELS.keys())
```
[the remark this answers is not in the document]
YasheshSavani: Or we can a have a method which creates a dictionary of available model on the system, so we don't have to check before running script that the model is available or not.

corneliusboehm:
This PR is completely contained in #142. We can test it now over there and apply potential final fixes.

PR #135, opened 2021-03-19
Use multi-processing and Flask-SocketIO for running the training

corneliusboehm:


corneliusboehm:
@YasheshSavani As discussed this seems to be working well. We should add a `Queue` for exchanging information between the processes and then this is good to go.

corneliusboehm (commented):
All working well :heavy_check_mark:

PR #136, opened 2021-03-19
Training module using SocketIO-Flask

YasheshSavani:
Description:

This PR will implement training module in sense studio by using `Flask-Socketio` for communication between web-page and back-end to show output on the implemented terminal.


tools/sense_studio/static/training.js
```diff
@@ -0,0 +1,43 @@
+// Ref: https://github.com/miguelgrinberg/Flask-SocketIO-Chat/blob/master/app/templates/chat.html
+var socket;
+$(document).ready(function () {
+
+    let project = $('#project').val();
+    let output_folder = $('#output_folder').val();
+
+    socket = io.connect('http://' + document.domain + ':' + location.port + '/train-model');
```
corneliusboehm, line 8: In our current case the `http://` is not even needed

tools/sense_studio/training.py
```diff
@@ -40,68 +33,88 @@ def train_model():
...
-#
-#     if os.path.exists(os.path.join(img_path, 'confusion_matrix.png')):
-#         return send_from_directory(img_path, 'confusion_matrix.png', as_attachment=True)
-#     return None
+@training_bp.route('/confusion-matrix/<string:project>', methods=['GET'])
+@training_bp.route('/confusion-matrix/<string:project>/<string:output_folder>', methods=['GET'])
+def confusion_matrix(project, output_folder):
+    path = utils.lookup_project_path(project)
+    img_path = os.path.join(path, output_folder, 'checkpoints')
+    return send_from_directory(img_path, filename='confusion_matrix.png', as_attachment=True)
```
corneliusboehm, line 120: I think this is not used

tools/sense_studio/training.py
```diff
@@ -40,68 +33,88 @@ def train_model():
...
+    global train_process
+    if train_process:
+        train_process.terminate()
+        train_process = None
+        message = "Training Cancelled."
+    return render_template('training.html', project=project, path=path, models=utils.BACKBONE_MODELS,
+                           is_disabled=False, message=message, output_folder=output_folder)
+
+
+@socketio.on('training_logs', namespace='/train-model')
```
corneliusboehm, line 84: Can we give this message another name? I took me some time to understand that this is the "connect" message from the client

tools/sense_studio/training.py
YasheshSavani: This is being used in `training.js` where the `img` tag is being created.

YasheshSavani: In that src will be given a path, which will render image using this method.

corneliusboehm: Oh, I saw the `url_for` call a few lines above just now

tools/sense_studio/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    from tools.sense_studio.annotation import annotation_bp
+    app.register_blueprint(annotation_bp, url_prefix='/annotation')
+
+    from tools.sense_studio.video_recording import video_recording_bp
+    app.register_blueprint(video_recording_bp, url_prefix='/video-recording')
+
+    from tools.sense_studio.training import training_bp
+    app.register_blueprint(training_bp, url_prefix='/training')
+
+    socketio.init_app(app, cors_allowed_origins="*")
```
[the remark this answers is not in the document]
YasheshSavani: I'll check it again why I added that

tools/sense_studio/training.py
YasheshSavani: Technically, it's an event name `training_logs` likewise `connect`. This one method works as connection, logs retriever, and at last disconnect sender.

YasheshSavani: yes, if it bothers more we can think about something else

corneliusboehm (approved):

PR #137, opened 2021-03-19
Check if ffmpeg is installed and leave error message

yoga-0125:
- Update function `syncRequest` in main.js by adding an optional argument for the HTTP request method (defalt `POST`)
- Update function `recordVideo` to send request to backend for `ffmpeg` installation check and leave an error mesage if it is not installed
- Add function `check_ffmpeg` in video_recording.py to check `ffmpeg` installation

### DIFF
--- a/tools/sense_studio/static/main.js
+++ b/tools/sense_studio/static/main.js
@@ -147,9 +147,11 @@ $.fn.form.settings.rules.uniqueClassName = function (className) {
 
 function syncRequest(url, data, method='POST') {
     let xhttp = new XMLHttpRequest();
-
     xhttp.open(method, url, false);
-    xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');
+
+    if (method == 'POST') {
+        xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');
+    }
 
     if (data) {
         xhttp.send(JSON.stringify(data));
--- a/tools/sense_studio/static/main.js
+++ b/tools/sense_studio/static/main.js
@@ -145,26 +145,10 @@ $.fn.form.settings.rules.uniqueClassName = function (className) {
 }
 
 
-function syncRequest(url, data) {
+function syncRequest(url, data, method='POST') {
     let xhttp = new XMLHttpRequest();
 
-    xhttp.open('POST', url, false);
-    xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');
-
-    if (data) {
-        xhttp.send(JSON.stringify(data));
-    } else {
-        xhttp.send();
-    }
-
-    return JSON.parse(xhttp.responseText);
-}
-
-
-function sendGetRequest(url, data) {
-    let xhttp = new XMLHttpRequest();
-
-    xhttp.open('GET', url, false);
+    xhttp.open(method, url, false);
     xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');
 
     if (data) {
--- a/tools/sense_studio/static/main.js
+++ b/tools/sense_studio/static/main.js
@@ -145,17 +145,15 @@ $.fn.form.settings.rules.uniqueClassName = function (className) {
 }
 
 
-function syncRequest(url, data, method='POST') {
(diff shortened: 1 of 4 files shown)


tools/sense_studio/video_recording.py
```diff
@@ -14,6 +15,12 @@
 video_recording_bp = Blueprint('video_recording_bp', __name__)
 
 
+@video_recording_bp.route('/ffmpeg-check', methods=['POST'])
```
YasheshSavani, line 18: The request method could have been `GET` as you are request data here.

YasheshSavani (changes requested):

tools/sense_studio/video_recording.py
YasheshSavani, line 18: The methods can be changed to `GET` as you are just requesting data here.

yoga-0125: > The methods can be changed to `GET` as you are just requesting data here.

Thanks and I update this point.

tools/sense_studio/static/main.js
```diff
@@ -161,6 +161,22 @@ function syncRequest(url, data) {
...
+    if (data) {
+        xhttp.send(JSON.stringify(data));
+    } else {
+        xhttp.send();
+    }
+
+    return JSON.parse(xhttp.responseText);
+}
+
+
```
YasheshSavani, line 179: Try to combine `sendGetRequest` and `syncRequest` into one method, What you can do here is pass one more parameter as function argument `method`, whose value will be `'GET'` or `'POST'` accordingly.

where for `xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');` you can use `method` as condition, because it is only needed to be sent when request type is `POST`.

YasheshSavani: Change accordingly in the code wherever `syncRequest` is used with appropriate method.

YasheshSavani: Add some description in the description box of PR.

tools/sense_studio/video_recording.py
```diff
@@ -14,6 +15,12 @@
 video_recording_bp = Blueprint('video_recording_bp', __name__)
 
 
+@video_recording_bp.route('/ffmpeg-check', methods=['GET'])
+def check_ffmpeg():
+    ffmpeg_installed = importlib.util.find_spec('ffmpeg')
```
corneliusboehm, line 20: I think what this checks is the availability of the python module `ffmpeg`. What we need to check however is if `ffmpeg` itself is installed system-wide.

tools/sense_studio/static/main.js
```diff
@@ -145,10 +145,10 @@ $.fn.form.settings.rules.uniqueClassName = function (className) {
 }
 
 
-function syncRequest(url, data) {
+function syncRequest(url, data, method='POST') {
     let xhttp = new XMLHttpRequest();
 
-    xhttp.open('POST', url, false);
+    xhttp.open(method, url, false);
     xhttp.setRequestHeader('Content-type', 'application/json; charset=utf-8');
```
YasheshSavani, line 152: `GET` request do not need RequestHeader, so here you can use method variable as condition.

tools/sense_studio/video_recording.py
```diff
@@ -14,6 +14,12 @@
 video_recording_bp = Blueprint('video_recording_bp', __name__)
 
 
+@video_recording_bp.route('/ffmpeg-check')
+def check_ffmpeg():
+    ffmpeg_installed = os.popen("dpkg-query -W -f='${Status}' ffmpeg").read()
```
corneliusboehm, line 19: I don't think we can assume that everyone is using `dpkg` (e.g. on a different OS). Can you maybe try running `ffmpeg -version` instead?

yoga-0125: > I don't think we can assume that everyone is using `dpkg` (e.g. on a different OS). Can you maybe try running `ffmpeg -version` instead?

update

corneliusboehm (approved):
Working well :+1:

PR #139, opened 2021-03-24
Split new project / import and use project name as folder name

corneliusboehm:
## Description

Before, it was confusing to have the same input fields and button for both creating new projects and importing existing ones. Now the two have been split and also redesigned to take less space per default.

Furthermore, you previously had to indicate the full path for the project to be created, while the project name was somehow unused. Now the project directory will automatically be created from the project name (with conversion to lower case and replacing spaces).

### Other Changes
- The new file `project_utils.py` has been created to bundle all project-related functions
- As part of switching to UIkit, the autocomplete mechanism for path search fields has been re-implemented by using [JavaScript-autoComplete](https://goodies.pixabay.com/javascript/auto-complete/demo.html)
- Form-checks have been added on the projects overview page

### DIFF
--- a/tools/sense_studio/annotation.py
+++ b/tools/sense_studio/annotation.py
@@ -20,6 +20,7 @@ from sklearn.linear_model import LogisticRegression
 from sense import SPLITS
 from sense.finetuning import compute_frames_features
 from tools import directories
+from tools.sense_studio import project_utils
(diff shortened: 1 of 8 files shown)


YasheshSavani (approved):
Tested and LGTM :+1:

PR #141, opened 2021-03-26
[Feature] Testing module in Sense Studio

YasheshSavani:
Description:

This PR will add a new feature of testing models in `sense-studio` to test custom trained models directly from the webpage by using two methods:

1. By giving, webcam feed as input
2. By giving path to an input video.

And also, an option to save the classified video for later use for both the above methods.


tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,113 @@
...
+                    Output folder ({{ project }}/...)
+                </label>
+                <div class="uk-form-controls">
+                    <div class="uk-inline uk-width-1-1">
+                        <span class="uk-form-icon" uk-icon="icon: folder"></span>
+                        <input class="uk-input" type="text" id="outputFolder" placeholder="Output Folder"
+                               autocomplete="off">
+                    </div>
+                </div>
+            </div>
```
YasheshSavani, line 63: I added this `output_folder` input taker to make saving video convenient for users. Otherwise, in traditional approach it will ask user to give whole path to the video save directory which I think was extra effort. 
So I divided original `path_out` argument from our old script into `output_folder` and `output_video_name` to make it easier.

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+        socket.emit('stream_video', {status: 'Socket Connected'});
+    });
+
+    socket.on('testing_images', function(message) {
+        streamVideo(message);
+    });
+
+    socket.on('success', function(message) {
+        if (message.status === 'Complete') {
+            addTerminalMessage('Stopping Inference...');
```
corneliusboehm, line 71: This message only shows up on the terminal a few seconds after pressing cancel when everything else is already stopped, so it's a little useless.

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+
+    await asyncRequest(url, data);
+
+    let socket = io.connect('/stream-video');
+    socket.on('connect', function() {
+        console.log('Socket Connected');
+        socket.emit('stream_video', {status: 'Socket Connected'});
+    });
+
+    socket.on('testing_images', function(message) {
```
corneliusboehm, line 65: How about `stream_frame` or `next_frame`?

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+            buttonCancelTest.disabled = true;
+        }
+    });
+
+    socket.on('testing_logs', function(message) {
+        addTerminalMessage(message.log);
+    });
+
+    buttonTest.disabled = true;
+    buttonCancelTest.disabled = false;
```
corneliusboehm, line 86: This is already done in lines 54/55

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+
+    document.getElementById('btnTest').disabled = false;
+    document.getElementById('btnCancelTest').disabled = true;
+}
+
+function toggleInputVideoField(){
+    let inputVideoDiv = document.getElementById('inputVideoDiv');
+    let inputSource = document.getElementsByName('inputSource');
+
+    for (let source of inputSource){
```
corneliusboehm, line 102: Do you actually need to fetch and read both sources? Doesn't one being checked automatically mean that the other one is not?

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+        disabled = true;
+    } else {
+        setFormWarning(pathLabel, pathInput, '');
+    }
+}
+
+async function startTesting(url){
+    let classifier = document.getElementById('classifier').value;
+    let inputVideoPath = document.getElementById('inputVideoPath').value;
+    let outputVideoName = document.getElementById('outputVideoName').value;
```
corneliusboehm, line 39: What happens if I enable "Save Output Video", enter a name in the field, and then disable it again? I think the video will still be saved, because we don't check the flag again here.

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+            inputVideoDiv.classList.add('uk-hidden');
+        } else if(source.value == 1 && source.checked){
+            inputVideoDiv.classList.remove('uk-hidden');
+        }
+    }
+}
+
+function toggleOutputVideoField(){
+    let outputVideoDiv = document.getElementById('outputVideoDiv');
+    let saveVideo = document.getElementsByName('saveVideo')[0];
```
corneliusboehm, line 113: Why not use an ID and have a simpler lookup?

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+
+{% block title %} Testing {% endblock %}
+
+{% block navigation %}
+    {% include 'navigation.html' %}
+{% endblock navigation %}
+
+{% block settings %}
+    {% include 'settings.html' %}
+{% endblock settings %}
```
corneliusboehm, line 11: There is no more `settings.html`

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+
+{% block main %}
+<div class="uk-container uk-container-xsmall">
+    <h1 class="uk-heading-medium uk-margin-large-top uk-margin-remove-bottom">Testing</h1>
+    <div class="uk-text-meta uk-text-large uk-margin-medium-bottom">
+        <span uk-icon="icon: folder"></span> {{ path }}
+    </div>
+    <form class="uk-form-stacked" method="post">
```
corneliusboehm, line 19: `method` is not necessary, because there is no submit action defined

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                            {% endfor %}
+                        {% else %}
+                            <option value="">No classifiers to select from</option>
+                        {% endif %}
+                    </select>
+                </div>
+            </div>
+        </div>
+
+        <div class="uk-child-width-1-2@s uk-grid-small" uk-grid>
```
corneliusboehm, line 40: If you define the width on the elements instead of via `child-width`, you can combine these grids

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+            <div>
+                <label class="uk-form-label" for="title">Window Title</label>
+                <div class="uk-form-controls">
+                    <div class="uk-inline uk-width-1-1">
+                        <span class="uk-form-icon" uk-icon="icon: pencil"></span>
+                        <input class="uk-input" type="text" id="title" placeholder="Title" autocomplete="off" />
+                    </div>
+                </div>
+            </div>
+            <div class="uk-margin-top">
```
corneliusboehm, line 50: Why this margin? Seems to mess up the design
image

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                            <option value="">No classifiers to select from</option>
+                        {% endif %}
+                    </select>
+                </div>
+            </div>
+        </div>
+
+        <div class="uk-child-width-1-2@s uk-grid-small" uk-grid>
+            <div>
+                <label class="uk-form-label" for="title">Window Title</label>
```
corneliusboehm, line 42: Didn't we say we can leave the `window title` out for now?

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                            &nbsp;&nbsp;Save Output Video
+                        </label>
+                     </div>
+                </div>
+            </div>
+        </div>
+
+        <div id="inputVideoDiv" class="uk-child-width-1-1@s uk-grid-small uk-hidden" uk-grid>
+            <div>
+                <label class="uk-form-label" for="inputVideoPath">Input video path</label>
```
corneliusboehm, line 82: Should we give the input video field an `.mp4` tag as well, because this is the only extension we currently look for?

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                               autocomplete="off" oninput="getPathInputs('inputVideoPathLabel', 'inputVideoPath');">
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div id="outputVideoDiv" class="uk-child-width-1-1@s uk-grid-small uk-hidden" uk-grid>
+             <div>
+                <label class="uk-form-label" for="outputVideoName">
+                    Output video name ({{ output_path_prefix }}..) <span class="uk-label">.mp4</span>
```
corneliusboehm, line 97: The order of fields is weird here, because if both optional fields are shown it becomes
1. Save video
2. Input video path
3. Output video path

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                        <input class="uk-input" type="text" id="outputVideoName" placeholder="Output video name"
+                               autocomplete="off">
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <button class="uk-button uk-button-primary uk-margin uk-margin-right" id="btnTest" type="button" tabindex="0"
+                onclick="startTesting('{{ url_for('.start_testing') }}');"
+                {% if not classifiers %} disabled {% endif %}>
```
corneliusboehm, line 111: This check doesn't seem to work

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+                onclick="cancelTesting('{{ url_for('.cancel_testing') }}');" disabled>
+            Cancel
+        </button>
+    </form>
+
+    <!--    Output video streaming frame-wise (Automatically filled) -->
+    <div id="videoStream" class="uk-margin uk-text-center">
+        <h4> Output Video Stream</h4>
+        <div class="uk-section uk-section-muted uk-padding-remove-vertical uk-inline">
+            <img id="frame" src="" height="480" width="640">
```
corneliusboehm, line 124: Can we remove the border around this box so that it looks similar to the one on the recording page?

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,136 @@
...
+            </div>
+        </div>
+
+        <div class="uk-child-width-1-1@s uk-grid-small" uk-grid>
+            <div>
+                 <div class="uk-form-controls">
+                     <div class="uk-width-1-1">
+                        <label>
+                            <input class="uk-checkbox" type="checkbox" name="saveVideo" onclick="toggleOutputVideoField();">
+                            &nbsp;&nbsp;Save Output Video
```
corneliusboehm, line 73: Spontaneous idea on the output video name: How about we just name it automatically, similar to how we do it in the video recording scripts. Then we don't need to fill an extra field.

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+
+@testing_bp.route('/<string:project>', methods=['GET'])
+def testing_page(project):
+    project = urllib.parse.unquote(project)
+    path = project_utils.lookup_project_path(project)
+    output_path_prefix = os.path.join(os.path.basename(path), 'output_videos', '')
+
+    default_checkpoint = os.path.join(path, 'checkpoints', 'best_classifier.checkpoint')
+    # If classifier checkpoint exist, get the path of sub-directory from checkpoints
+    classifiers = [os.path.join('checkpoints', os.path.basename(d)) for d in glob.glob(f"{path}/checkpoints/*")
```
corneliusboehm, line 38: I would suggest `os.listdir` instead of `glob` here

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+                   if os.path.exists(os.path.join(d, 'best_classifier.checkpoint'))
+                   and os.path.isdir(d)]
+    classifiers.extend(['checkpoints/'] if os.path.exists(default_checkpoint) else [])
```
corneliusboehm, line 41: How about this to be more straightforward?
```python
if os.path.exists(default_checkpoint)
    classifiers.apend('checkpoints/')
```

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return render_template('testing.html', project=project, path=path, output_path_prefix=output_path_prefix,
+                           classifiers=classifiers)
+
+
+@testing_bp.route('/start-testing', methods=['POST'])
+def start_testing():
+    data = request.json
+    path_in = data['inputVideoPath'] or ''
+    path_out = data['outputVideoName'] or ''
```
corneliusboehm, line 52: Are the `or`s even necessary?

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+def cancel_testing():
+    global test_process
+    global queue_signal
+    if test_process:
+        # Send signal to stop inference
+        queue_signal.put(-1)
+
+        # Wait until process is complete
+        test_process.join()
+        test_process.terminate()
```
corneliusboehm, line 98: It shouldn't be necessary to call `terminate` after `join`, except you provide a `timeout`

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+            try:
+                output = queue_testing_output.get(timeout=1)
+                if type(output) == str:
+                    emit('testing_logs', {'log': output})
+                else:
+                    # Encode frame as jpeg
+                    frame = cv2.imencode('.jpg', output)[1].tobytes()
+                    # Encode frame in base64 version and remove utf-8 encoding
+                    frame = base64.b64encode(frame).decode('utf-8')
+                    emit('testing_images', {'image': f"data:image/jpeg;base64,{frame}"})
```
corneliusboehm, line 120: On the training page I did something similar, but on the javascript side. I would say we should treat both cases similar, but I don't mind which version we use.

tools/sense_studio/testing.py
```diff
@@ -0,0 +1,133 @@
...
+
+        test_process.terminate()
+        test_process = None
+    except AttributeError:
+        # test_process has been cancelled and is None
+        pass
+    finally:
+        queue_testing_output.close()
+
+    emit('success', {'status': 'Complete', 'image': ""})
```
corneliusboehm, line 133: Can we just remove the `image` argument here?

YasheshSavani: I wanted that so at the end the video stream placeholder becomes empty otherwise it will stick at last frame.

tools/sense_studio/sense_studio.py
```diff
@@ -101,8 +103,10 @@ def browse_directory():
     project = data['project']
 
     subdirs = [d for d in glob.glob(f'{path}*') if os.path.isdir(d)] if os.path.isabs(path) else []
-    full_path = os.path.join(path, project_utils.get_folder_name_for_project(project))
+    subfiles = [f for f in glob.glob(f'{path}*.mp4') if os.path.isfile(f)] if os.path.isabs(path) else []
+    subfiles.extend(subdirs)
```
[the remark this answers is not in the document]
YasheshSavani: we can do that

tools/sense_studio/static/testing.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+function streamVideo(message) {
+    let frame = document.getElementById('frame');
+    frame.src = message.image;
+}
+
+async function getPathInputs(pathLabelId, pathInputId) {
+    let pathLabel = document.getElementById(pathLabelId);
+    let pathInput = document.getElementById(pathInputId);
```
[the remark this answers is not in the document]
YasheshSavani: I kept it as I thought this might be used in future but yes, we can do that.

tools/sense_studio/static/testing.js
YasheshSavani: To see which one is checked we need to go through all, as far as I have searched about it

tools/sense_studio/templates/testing.html
YasheshSavani: yes, but I could not come to a proper design without that :sweat_smile:

tools/sense_studio/static/testing.js
YasheshSavani: What if on toggle back we clear the input field? This case will be same for the input source radio button.

tools/sense_studio/static/testing.js
YasheshSavani: Now, I understood your point and have done changes accordingly.

tools/sense_studio/templates/testing.html
```diff
@@ -0,0 +1,124 @@
...
+                               autocomplete="off" oninput="editInputVideoPath();">
+                    </div>
+                </div>
+            </div>
+
+            <div>
+                 <div class="uk-form-controls uk-width-1-1">
+                    <input class="uk-checkbox" type="checkbox" id="saveVideo" autocomplete="off"
+                           onclick="toggleOutputVideoField();">
+                        Save Output Video
```
corneliusboehm, line 80: Neither the `Input Video` nor the `Save Output Video` field force you to actually enter the required information. So it is possible to check these boxes and start a training, which will ignore those settings. It is even possible to start a training with an incomplete or wrong input video path, which will just stop without any visible error message.
I think this is the final missing detail on this page.

corneliusboehm (commented):
All looking good and working really well. Just a final comment left to make the interface safe against users :grimacing:

tools/sense_studio/sense_studio.py
```diff
@@ -105,6 +107,7 @@ def browse_directory():
     project_dir = project_utils.get_folder_name_for_project(project)
     full_path = os.path.join(path, project_dir)
 
+    video_files = [f for f in glob.glob(f'{path}*.mp4') if os.path.isfile(f)] if os.path.isabs(path) else []
```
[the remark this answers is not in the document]
YasheshSavani: Not needed, I just checked so removed it

tools/sense_studio/templates/testing.html
YasheshSavani: did updates to handle  these situations

tools/sense_studio/static/testing.js
```diff
@@ -36,18 +36,32 @@ async function startTesting(url) {
...
     let path = document.getElementById('path').value;
     let title = document.getElementById('title').value;
     let buttonTest = document.getElementById('btnTest');
     let buttonCancelTest = document.getElementById('btnCancelTest');
     let frame = document.getElementById('frame');
 
+    if (!inputVideoPathValue || !inputVideoPathValue.endsWith('.mp4')) {
+        setFormWarning(inputVideoPathLabel, inputVideoPath, 'Please provide a valid .mp4 file');
+        return false;
+    }
```
corneliusboehm, line 54: This part currently doesn't allow to stream from the webcam, because in that case the `inputVideoPathValue` is empty

tools/sense_studio/static/testing.js
```diff
@@ -36,18 +36,32 @@ async function startTesting(url) {
...
+
+    if (saveVideo.checked && !outputVideoNameValue) {
+        setFormWarning(outputVideoNameLabel, outputVideoName, 'Please provide a video name');
+        return false;
+    }
```
corneliusboehm, line 59: This warning is never reset, so it will even stay when a name has been provided

corneliusboehm (changes requested):
There are a few issues with this solution, as described in the comments.
Also on the other pages we usually check those conditions already on input and disable/enable the action button accordingly. Would be nice to keep the behavior similar here.

tools/sense_studio/static/testing.js
YasheshSavani: > Also on the other pages we usually check those conditions already on input and disable/enable the action button accordingly. Would be nice to keep the behavior similar here.

This is too complicated as both the event `input_video` and `save_output_video` are independent of each other so if they are being manipulated individually the solution of doing everything in `oninput` is good. But, if they are both selected together then the functionality created for both individually will make it complicated.

I tried many ways to solve this but could not if we want have it in `oninput` event listeners.

In simple terms, two processes racing for a resource in parallel but there is no lock system.

corneliusboehm: I am processing all conditions at the same time now, which should solve the problem

corneliusboehm (approved):

PR #146, opened 2021-04-01
Fix confusion matrix display

corneliusboehm:
This PR fixes two error cases related to the display of the confusion matrix:

### 1. Caching
The browser was apparently caching the image of the confusion matrix (given through a URL), so if you ran a training multiple times in the same directory, you would always only see the first confusion matrix.
Now we directly send the binary image in the response from the server, so no URL can be cached.

### 2. Showing confusion matrix from previous run
If you ran another training in an existing checkpoint directory, but cancelled it before the first epoch (or ran a temporal training that doesn't produce a confusion matrix), the interface would still show the confusion matrix from the previous training run.
Now we use an `Event` to check, if a new confusion matrix has actually been produced in this training run.

### DIFF
--- a/tools/sense_studio/training.py
+++ b/tools/sense_studio/training.py
@@ -119,5 +119,7 @@ def send_training_logs(msg):
         img_base64 = base64.b64encode(data)
         if img_base64:
             emit('success', {'status': 'Complete', 'img': img_base64})
+        else:
+            emit('failed', {'status': 'Failed'})
     else:
         emit('failed', {'status': 'Failed'})
--- a/tools/sense_studio/training.py
+++ b/tools/sense_studio/training.py
@@ -1,3 +1,4 @@
+import base64
 import multiprocessing
 import os
 import queue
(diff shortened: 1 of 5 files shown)


tools/sense_studio/training.py
```diff
@@ -106,20 +113,11 @@ def send_training_logs(msg):
...
-    if os.path.exists(img_path):
-        img_path = url_for('training_bp.confusion_matrix',
-                           project=msg['project'],
-                           output_folder=msg['outputFolder'])
-        emit('success', {'status': 'Complete', 'img_path': img_path})
+    if confmat_event.is_set() and os.path.exists(img_path):
+        with open(img_path, 'rb') as f:
+            data = f.read()
+        img_base64 = base64.b64encode(data)
+        if img_base64:
```
YasheshSavani, line 120: I put this because I did some testing on this. This will cause failure in case the image is being written on drive by process and we cancel it in between which will result in empty text file in place of image. Which on web-page will show broken image icon that will cause problem in user understanding it. So this condition ensures that if the file is written completely or not and if yes, then send it.

corneliusboehm: Thanks for the fix! However, we should also send the `failed` response in this case.

YasheshSavani (approved):
Working fine :+1:

PR #147, opened 2021-04-01
[Fix] First training does not start

YasheshSavani:
Description:

This PR will fix the issue of `OSError: handle is closed` which occurs because of unavailability of the training process object in the request done by socketio.


corneliusboehm (approved):
Working fine. Let's hope this fixes the issue :grin:

PR #148, opened 2021-04-06
Check available models

yoga-0125:
- Add function `model_exist()` in class `ModelConfig` in `loading.py` to check if a model exists 
- Update `SUPPORTED_MODEL_CONFIGURATIONS` in `utils.py `
- Edit this branch on the top of branch `feature/threading-training-module`, but it seems to disappear


sense/loading.py
```diff
@@ -55,6 +55,16 @@ def __init__(self, model_name: str, version: str, feature_converters: List[str])
 
         self.feature_converters = feature_converters
 
+    def model_exist(self):
+        model_weights = MODELS[self.model_name][self.version]
+        path_weights = {name: model_weights[name] for name in ['backbone'] + self.feature_converters}
+        files_exist = all(os.path.exists(prepend_resources_path(path)) for path in path_weights.values())
```
corneliusboehm, line 61: This code looks almost identical to the checks below in `get_weights`. Can you manage to reuse it there?

tools/sense_studio/utils.py
```diff
@@ -12,13 +12,15 @@
...
 
-SUPPORTED_MODEL_CONFIGURATIONS = [
+ALL_SUPPORTED_MODEL_CONFIGURATIONS = [
     ModelConfig('StridedInflatedEfficientNet', 'pro', []),
     ModelConfig('StridedInflatedMobileNetV2', 'pro', []),
     ModelConfig('StridedInflatedEfficientNet', 'lite', []),
     ModelConfig('StridedInflatedMobileNetV2', 'lite', []),
 ]
 
+SUPPORTED_MODEL_CONFIGURATIONS = [i for i in ALL_SUPPORTED_MODEL_CONFIGURATIONS if i.model_exist() is True]
```
corneliusboehm, line 22: This is only evaluated once when the python stack is started. I would prefer to check that live whenever the list of models is requested, because a user might have downloaded the models in the meantime.

corneliusboehm (approved):
Working well :+1: We should only make sure to add better handling for the case where no model is available on the training page in the future.

PR #151, opened 2021-04-09
[Fix] Lab2int fix for temporal training

YasheshSavani:
Description:

This PR will fix issue of wrong tags names in Lab2Int file for temporal training. 

Issue: 
It was not taking updated tag names of the classes in temporal training. Instead, it was taking `{class_name}_position_1` and `{class_name}_position_2` as tag names.

Update from Guillaume's commit on his branch `feature/ui-kit-smart-tv`.
Link: https://github.com/TwentyBN/sense/commit/fb323f21cee00cac6b8c793d0838763fbf00c10a

### DIFF
--- a/sense/finetuning.py
+++ b/sense/finetuning.py
@@ -18,7 +18,6 @@ from sense.engine import InferenceEngine
 from sense.utils import clean_pipe_state_dict_key
 from tools import directories
 from tools.sense_studio import utils
-from tools.sense_studio.project_utils import PROJECT_CONFIG_FILE
 
 MODEL_TEMPORAL_DEPENDENCY = 45
 MODEL_TEMPORAL_STRIDE = 4
--- a/sense/finetuning.py
+++ b/sense/finetuning.py
@@ -97,7 +97,7 @@ class FeaturesDataset(torch.utils.data.Dataset):
         return [features, self.labels[idx], temporal_annotation]
 
 
-def generate_data_loader(features_dir, tags_dir, label_names, label2int,
+def generate_data_loader(project_config, features_dir, tags_dir, label_names, label2int,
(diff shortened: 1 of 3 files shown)


tools/train_classifier.py
```diff
@@ -124,26 +124,30 @@ def train_model(path_in, path_out, model_name, model_version, num_layers_to_fine
     # Find label names
     label_names = os.listdir(directories.get_videos_dir(path_in, 'train'))
     label_names = [x for x in label_names if not x.startswith('.')]
-    label_counting = ['counting_background']
 
-    for label in label_names:
-        label_counting += [f'{label}_position_1', f'{label}_position_2']
+    with open(os.path.join(path_in, 'project_config.json')) as f:
+        project_config = json.load(f)
```
corneliusboehm, line 129: This function can still be called as a stand-alone script, so the `project_config.json` doesn't necessarily exist.
Also we have a `PROJECT_CONFIG_FILE` constant that could be used here.

sense/finetuning.py
```diff
@@ -97,7 +97,7 @@ def __getitem__(self, idx):
         return [features, self.labels[idx], temporal_annotation]
 
 
-def generate_data_loader(features_dir, tags_dir, label_names, label2int,
+def generate_data_loader(project_config, features_dir, tags_dir, label_names, label2int,
```
corneliusboehm, line 100: Side note: The proper way to attribute these changes to @guillaumebrg and keep version history clean would have been to `cherry-pick` his commit. Did you try that?

YasheshSavani: I specifically did not `cherry-pick` it, I used `checkout` to get the files from his branch.

corneliusboehm (approved):

PR #152, opened 2021-04-12
[Fix] Train model button disabled if no models exist

YasheshSavani:
Description;

This PR will fix the issue of `Train Model` button staying active even if the models to choose from are not available on user's system.


yoga-0125:
looks good for me.

yoga-0125 (approved):

corneliusboehm:
This works, but it doesn't give the user any clue, why training is not possible. I would suggest adding the error color to the model selection field and showing a hint somewhere that they should go to 20bn.com to download the weights.

PR #154, opened 2021-04-12
[Feature] Clarify num of layers to finetune

YasheshSavani:
Description:

This PR will update the training page in sense studio. Below are the changes:

1. It will replace the original input taker for `Number of layers to finetune` with a slider and an output tag beside it for showing the selected layers.
2. The output tag will show `Classification Layer` when selected number of layers to finetune is `0`.
3. It will set default `numberOfLayersToFinetune` value based on the selected model.


corneliusboehm (approved):
I made some updates to add the correct `max` value per model type and to always refresh the slider on page load.

PR #155, opened 2021-04-13
Allow video to be recorded with original framerate in run_custom_classifier

yoga-0125:
- resample frames of video in function `run_inference()` in controller.py
- add function `get_frames()` in camera.py to extract frames number

### DIFF
--- a/tools/run_custom_classifier.py
+++ b/tools/run_custom_classifier.py
@@ -31,7 +31,6 @@ from sense.downstream_tasks.postprocess import PostprocessClassificationOutput
 from sense.loading import build_backbone_network
 from sense.loading import load_backbone_model_from_config
 from sense.loading import update_backbone_weights
-from sense.camera import VideoSource
 
 
 if __name__ == "__main__":
@@ -67,9 +66,6 @@ if __name__ == "__main__":
 
     # Concatenate feature extractor and met converter
     net = Pipe(backbone_network, gesture_classifier)
-    video = VideoSource(filename=path_in, camera_id=camera_id)
-    net.feature_extractor.fps = video.get_fps()
-    video.release()
 
     postprocessor = [
         PostprocessClassificationOutput(INT2LAB, smoothing=4)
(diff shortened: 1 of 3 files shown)


tools/run_custom_classifier.py
```diff
@@ -66,6 +67,9 @@
 
     # Concatenate feature extractor and met converter
     net = Pipe(backbone_network, gesture_classifier)
+    video = VideoSource(filename=path_in, camera_id=camera_id)
+    net.feature_extractor.fps = video.get_fps()
+    video.release()
```
corneliusboehm, line 72: Unfortunately, I think this is the opposite of what we need. Here you are adapting the model fps to the one of the video, but we actually need to keep the model fps the same as in training while reading the video at the correct frame rate.

Something similar is already done during training, as you can see in `finetuning.py:extract_frames()`:
```python
    # Read frames from video
    video_source = camera.VideoSource(size=inference_engine.expected_frame_size, filename=video_path)
    video_fps = video_source.get_fps()
    frames = []

    while True:
        images = video_source.get_image()
        if images is None:
            break
        else:
            image, image_rescaled = images
            frames.append(image_rescaled)

    frames = uniform_frame_sample(np.array(frames), inference_engine.fps / video_fps)
```

On another branch, @antoinemrcr has already expanded the `uniform_frame_sample` method to also handle upsampling:
```python
def uniform_frame_sample(video, sample_rate):
    """
    Uniformly sample video frames according to the provided sample_rate.
    """
    depth = video.shape[0]
    if sample_rate < 1.:
        indices = np.arange(0, depth, 1. / sample_rate)
        offset = int((depth - indices[-1]) / 2)
        sampled_frames = (indices + offset).astype(np.int32)
        return video[sampled_frames]
    elif sample_rate > 1:
        num_frames = int(sample_rate * depth)
        new_indices = np.array(list(range(num_frames)))
        new_indices = (new_indices / sample_rate).astype(int)
        new_indices = np.clip(new_indices, 0, depth -1)
        video = video[new_indices, :]
    return video
```

PR #156, opened 2021-04-15
[Feature] Clarify path for new project

YasheshSavani:
Description:

This PR will increase the size of the new project creation card and will add one more disabled field beside the project path to show the project name for which the project folder will be created.


guillaumebrg (commented):
Made a few suggestions here: #157 
Can you have a look? Let's jump on a call then

guillaumebrg (approved):
LGTM!

PR #157, opened 2021-04-15
Suggestions about "[Feature] Clarify path for new project"

guillaumebrg:
@YasheshSavani Can you try this and let me know what you think?


YasheshSavani (commented):
LGTM, but let's have it as an another option with my solution. What do you say?

PR #159, opened 2021-04-16
Safe path handling in the full path display

corneliusboehm:
Thanks for the design update in #156!

There were a few small issues with the previous solution though:
- The project name was shown as the project directory, but actually the name is transformed on the backend before the directory is created (e.g. to avoid spaces in the folder name)
- The shown path would be incorrect on Windows, because it would always append a `/`
- When clearing both input fields, the label would still show a single '/'

Now the proper path handling is done on the backend and all necessary information is transferred to the JS side.


YasheshSavani (approved):
LGTM,  Thanks for the changes :+1:

PR #160, opened 2021-04-19
Resample testing videos

corneliusboehm:
Make sure that videos during testing are also presented to the model at the correct frame rate. Currently, they would be read at their original frame rate while assuming it to be the model frame rate, making them look slower or faster.

- The `uniform_frame_sample` method has been extended to also handle the case of upsampling
- A `VideoSource` can now have a `target_fps` defined. If this is given, the video is immediately read and resampled on initialization.
- This same mechanism is now used in both training and testing
- Unit tests for the new behavior have been added

This is a different approach to #155

### DIFF
--- a/sense/camera.py
+++ b/sense/camera.py
@@ -8,6 +8,25 @@ from typing import Optional
 from typing import Tuple
 
 
+def uniform_frame_sample(video, sample_rate):
+    """
+    Uniformly sample video frames according to the provided sample_rate.
+    """
+    available_frames = video.shape[0]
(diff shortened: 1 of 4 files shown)


yoga-0125:
Looks great for me. Only one question is that the unit test part for `TestVideoStream` and `TestVideoWriter` and `if __name__ == '__main__': unittest.main()` are removed?

yoga-0125 (approved):

guillaumebrg (approved):
Looks good to me! :fire:

PR #161, opened 2021-04-21
[Feature] Make tags class independent

YasheshSavani:
Description:

This PR will change the annotation tag creation feature from class dependent tags to class independent tags (project wide tags).

Features:

1. Create tag for project wide use
2. Update, delete features for the tag
3. New method to select multiple tags for classes 
4. Reuse any tag for different classes


corneliusboehm (changes requested):
I've done a first review now. The UI looks very nice, good job there. However I have still observed a few issues, which I'm already working on solving:
- Backwards compatibility: Transformation of previous tags and annotations into new scheme
- Handling of `background` tag: Needs to be added manually to all classes, can currently be removed, is not added to old project configs
- Gaps in tag indices: The tag indices are used for building `label2int_temporal_annotation` and those indices are later used to get the correct element from the model's output vector. Now if the tag indices have gaps (because a tag was removed), our indexing into the output vector will be broken.

corneliusboehm (approved):

PR #162, opened 2021-04-27
Updated sense studio wiki

sunny-panchal:
Basic wiki update. Feel free to make some changes!


docs/wiki/sensestudio.md
```diff
@@ -0,0 +1,87 @@
...
+---
+
+## Creating New Classes
+
+New classes can be added from the `Project Details` tab. Creating a
+new class automatically updates the video folder structure and project config. 
+
+**NOTE**: Class names can be edited directly on SenseStudio from the `Project Details` page. 
+Updating it in `project_config.json` may break SenseStudio if the relevant folder names are not
+updated.
```
corneliusboehm, line 39: Not sure if we should even hint at the possibility of manually editing `project_config.json` :grimacing:

docs/wiki/sensestudio.md
```diff
@@ -0,0 +1,87 @@
...
+Temporal annotations can be turned on in the settings window in the top-right corner
+of the `Project Details` page. Once enabled, each class card will expand, and you will see a list of
+tags, and a summary of the number of available videos that have been annotated. For now, 2 tags 
+can be defined in addition to a default background class. Names can be edited by clicking on the edit icon 
+on the card. 
+
+When ready to annotate your videos, click on the "Annotate" button. This will automatically process
+the available videos to:
+1. Extract single frames from your videos 
+2. Precompute features on those frames
```
corneliusboehm, line 75: We are not explaining yet, why features would be needed for this. Also, assisted tagging is disabled by default.

docs/wiki/sensestudio.md
```diff
@@ -0,0 +1,87 @@
...
+
+You can then click on the video you want to annotate and be presented with a grid-view of each 
+frame where the appropriate tag can be selected. Remember to click `Submit` at the bottom of the 
+page when you are finished to return to the previous page.
+
+**NOTE:** The model will perform best on annotations that occur after the first 2-3 seconds of the video
```
corneliusboehm, line 81: Should we rephrase this to make clear that the model only might have problems during the first 2-3 seconds, while all later annotations are fine? Currently, this might be mistaken as only the tags at about 2-3s being learned well.

corneliusboehm (approved):
This is a great starting point for SenseStudio documentation, thanks for adding that! We should add screenshots soon, but for now this is good to go.

corneliusboehm:
As of just now we also have a Testing module :grin:

docs/wiki/sensestudio.md
sunny-panchal: We can remove the reference :p

docs/wiki/sensestudio.md
sunny-panchal: removed for now

corneliusboehm (approved):
Thanks for the updates!

PR #163, opened 2021-04-27
add confusion matrix for temporal training and sort labels name alphabetically

yoga-0125:
- sort `label_names` in `train_classifier.py`
- generate confusion matrix for temporal training in `finetuning.py`

### DIFF
--- a/sense/finetuning.py
+++ b/sense/finetuning.py
@@ -290,9 +290,8 @@ def extract_features(path_in, model_config, net, num_layers_finetune, use_gpu, n
         log_fn('\n')
 
 
-def training_loops(net, train_loader, valid_loader, use_gpu, num_epochs, lr_schedule, label_names,
-                   label_names_temporal_cm, path_out, temporal_annotation_training=False, log_fn=print,
-                   confmat_event=None):
+def training_loops(net, train_loader, valid_loader, use_gpu, num_epochs, lr_schedule, label_names, label_names_temporal,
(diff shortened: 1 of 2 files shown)


tools/train_classifier.py
```diff
@@ -124,6 +124,7 @@ def train_model(path_in, path_out, model_name, model_version, num_layers_to_fine
 
     # Find label names
     label_names = os.listdir(directories.get_videos_dir(path_in, 'train'))
+    label_names.sort()
```
YasheshSavani, line 127: if possible  use `natural sorting` instead of basic sorting.  For that  we have a  library called `natsort`. You can find that in `testing.py` on line 49.

sense/finetuning.py
```diff
@@ -399,7 +400,8 @@ def run_epoch(data_loader, net, criterion, optimizer=None, use_gpu=False,
     top1 = np.mean(epoch_labels == epoch_top_predictions)
     loss = running_loss / len(data_loader)
 
-    cnf_matrix = confusion_matrix(epoch_labels, epoch_top_predictions)
+    cnf_matrix = confusion_matrix(epoch_labels, epoch_top_predictions,
+                                  labels=range(min(epoch_labels), max(epoch_labels) + 1))
```
corneliusboehm, line 404: Would it be safer to go from 0 to `len(label_names_temporal)` just in case the epoch doesn't contain an example for one of those endpoints?

corneliusboehm (approved):
Working well :+1: Next we should see if we can use a better metric than accuracy for event detection.

PR #165, opened 2021-04-29
[FIX] Incorrect equality check in test

sunny-panchal:
Sorry, that was my mistake. This should fix it.


tests/test_camera.py
```diff
@@ -106,7 +106,10 @@ def test_write(self):
             _, frame_out = output_video.read()
             if not ret:
                 break
-            assert np.all([frame_in, frame_out])
+            assert np.ndarray([frame_in == frame_out]).all()
```
corneliusboehm, line 109: When I run the following snippet, it fails with the shown error:
```python
>>> frame_in = np.random.random((2,3))
>>> frame_out = frame_in
>>> np.ndarray([frame_in == frame_out])
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: only integer scalar arrays can be converted to a scalar index
```

I think the proper way to write this check would be
```python
assert (frame_in == frame_out).all()
```

However, this hints at the problem that our test case apparently never reached that line. I guess this is because `input_video` reader is already released above, so it can no longer be read.

corneliusboehm (changes requested):

tests/test_camera.py
sunny-panchal: I see, also turns out that the written video isn't identical. Visually it is but the re-encoding introduces some variance. Should we bother with the equivalence check?

tests/test_camera.py
```diff
@@ -99,14 +99,16 @@ def test_write(self):
...
+        input_video = cv2.VideoCapture(VIDEO_PATH)
 
         # Check whether new output file is the same as the original
         while True:
             ret, frame_in = input_video.read()
             _, frame_out = output_video.read()
             if not ret:
                 break
-            assert np.ndarray([frame_in == frame_out]).all()
+            assert (np.isclose(frame_in, frame_out, atol=50)).all()
```
corneliusboehm, line 111: We could also use [`np.allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html#numpy.allclose)

tests/test_camera.py
```diff
@@ -99,14 +99,16 @@ def test_write(self):
         input_video.release()
```
corneliusboehm, line 99: Why did `test_video.mp4` change in this update? Is it maybe rewritten anywhere in the unit tests?

tests/test_camera.py
sunny-panchal: Das is richtig :facepalm:

tests/test_camera.py
sunny-panchal: The writer output a video using a different encoder, resulting in some minor variation,  so I replaced the input video with the one created to match the encoding style

corneliusboehm: Makes sense. So now there should be no more changes after running the tests, right?

corneliusboehm (approved):

PR #166, opened 2021-05-04
Add general changes from volleyball project

corneliusboehm:
As part of the volleyball example in https://github.com/corneliusboehm/sense/pull/5, some general changes were added, which would make sense to be contained in this repository's main branch.

### Features
- `DisplayCounts` has been added as a generalized version of `DisplayRepCounts`, which doesn't show the `Exercise:` prefix and can highlight counts that have recently been updated.
- `title` in `DisplayResults` has been made optional
- The class `AggregatedPostProcessors` has been added, which takes a list of `PostProcessors` and stores their results together in the output directory under a given `out_key`. This class replaces the previous `PostprocessRepCounts` and `PostprocessEventCounts`.
- The class `ExerciseSpecificRepCounter` has been generalized to `TwoPositionsCounter`
- `EventCounter` can now take a custom `out_key`
- In `get_relevant_weights` you can now request a specific feature converter

### Bug Fixes
- If the `--path_in` file didn't exist, `cv2.VideoCapture` was still created and did not throw an error except that the reported fps was 0. Now a `FileNotFoundError` will be raised.
- If the setup of the stack took a little longer (e.g. because a long video needed to be read and resampled), the displayed video and model fps would start very low and only slowly recover, because the first measured delta included the setup time. Now there is an `initialize` call on `BaseDisplay` which will be called when all setup is done.
- Add a check if `temporal_annotation` is empty


examples/run_fitness_rep_counter.py
```diff
@@ -24,11 +24,12 @@
 
 import sense.display
 from sense.controller import Controller
-from sense.downstream_tasks.fitness_rep_counting import INT2LAB
+from sense.downstream_tasks.fitness_rep_counting import INT2LAB, LAB2INT
```
guillaumebrg, line 27: Stick to one import per line?

sense/display.py
```diff
@@ -160,8 +195,13 @@ def __init__(
...
         self.running_delta_time_camera = 1. / expected_camera_fps if expected_camera_fps else 0
-        self.last_update_time_camera = time.perf_counter()
-        self.last_update_time_inference = time.perf_counter()
+        self.last_update_time_camera = None
+        self.last_update_time_inference = None
+
+    def initialize(self):
+        now = time.perf_counter()
+        self.last_update_time_camera = now
+        self.last_update_time_inference = now
```
guillaumebrg, line 204: :+1: 

The slow ramp up phase was annoying me too

guillaumebrg (approved):
LGTM! Just made one small comment.

sense/finetuning.py
```diff
@@ -92,8 +92,8 @@ def __getitem__(self, idx):
                 position = np.random.randint(minimum_position, num_preds - self.num_timesteps)
                 features = features[position: position + self.num_timesteps]
             # will assume that we need only one output
-        if temporal_annotation is None:
-            temporal_annotation = [-100]
+        if temporal_annotation is None or len(temporal_annotation) == 0:
+            temporal_annotation = np.array([-100])
```
guillaumebrg, line 96: Why was adding `len(temporal_annotation) == 0` needed here? When does that happen?
Will this lead to ignore videos that were annotated but didn't contain non-background tags?

guillaumebrg: > Will this lead to ignore videos that were annotated but didn't contain non-background tags?

I guess not... Had a quick look and I think `temporal_annotation` is full of 0s in that case. If that's correct, then you can ignore my comment above ;)

corneliusboehm: When committing this recently I also wasn't sure, which case this was added for, but at the time of the hackathon I'm pretty sure it crashed exactly there. Leaving it in shouldn't hurt, right? :shrug:

PR #167, opened 2021-05-11
Backwards compatibility and missing updates for project tags

corneliusboehm:
### Features
- Add backwards compatibility update when loading project config
  - Setup dictionary of project tags, update class tags and translate existing annotations
  - Check for all other project settings that have been added later
- Treat `background` as persistent default label
- Correctly handle tag indices that might have gaps (when tags have been removed), especially when building `label2int` dictionary
- Save `max_tag_index` so that a removed tag can not be overwritten by a new one
- Reset annotations for tags that have been removed
- Let `logreg` work on an arbitrary number of tags

### Smaller Changes
- Don't show "This tag already exists" when leaving a tag name unchanged
- Flip tag name and index in project tags dictionary for easier processing
- Sort tags by value
- Remove duplicate `FileNotFoundErrors`
- Move buttons on same line with edit field
image
- Reset class name when editing is cancelled (similar to how it was done for tags)
- Directly send submitted forms for saving/removing project tags to the backend

### DIFF
--- a/tools/sense_studio/project_utils.py
+++ b/tools/sense_studio/project_utils.py
@@ -31,11 +31,62 @@ def lookup_project_path(project_name):
     return projects[project_name]['path']
 
 
+def _backwards_compatibility_update(path, config):
+    updated = False
+
+    if 'use_gpu' not in config:
+        config['use_gpu'] = False
+        updated = True
+
+    if 'temporal' not in config:
+        config['temporal'] = False
+        updated = True
+
+    if 'assisted_tagging' not in config:
+        config['assisted_tagging'] = False
+        updated = True
+
+    if 'video_recording' not in config:
+        config['video_recording'] = {
+            'countdown': 3,
+            'recording': 5,
+        }
+        updated = True
+
+    if 'project_tags' not in config:
+        # Collect class-wise tags
+        old_classes = config['classes']
+        project_tags_list = []
+        for class_name, class_tags in old_classes.items():
+            project_tags_list.extend(class_tags)
+
+        # Assign project-wide unique indices to tags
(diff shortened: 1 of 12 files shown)


tools/sense_studio/project_utils.py
```diff
@@ -31,11 +31,89 @@ def lookup_project_path(project_name):
...
+
+    if 'project_tags' not in config:
+        # Collect class-wise tags
+        old_classes = config['classes']
+        project_tags_list = []
+        for class_name, class_tags in old_classes.items():
+            project_tags_list.extend(class_tags)
+
+        # Assign project-wide unique indices to tags (0 is reserved for 'background')
+        project_tags = {idx + 1: tag_name for idx, tag_name in enumerate(project_tags_list)}
```
YasheshSavani, line 64: Maybe `sort` class labels first and then assign them index? What do you say? @corneliusboehm

corneliusboehm: Sure, why not

tools/sense_studio/static/main.js
```diff
@@ -398,20 +400,23 @@ async function deselectTagFromList(classIdx, tagIndex, tagName, path, className)
...
     let disabled = false;
 
     if (tag.value === '') {
         setFormWarning(errorLabel, tag, '');
         disabled = true;
-    } else if (projectTagNames.includes(tag.value)) {
+    } else if (tag.value !== originalTagName && projectTagNames.includes(tag.value)) {
         setFormWarning(errorLabel, tag, 'This tag name already exist');
         disabled = true;
+    } else if (tag.value === 'background') {
```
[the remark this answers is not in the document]
corneliusboehm: Good point. I'm using `toLowerCase()` now to catch all such cases.

PR #168, opened 2021-05-11
use classes from project config if config exists

YasheshSavani:
Description:

This PR will add functionality to use classes from project config-if project config exists else use directory names as class labels in training.


corneliusboehm (commented):
Works :heavy_check_mark: I just noticed that in the `extract_features` call above, the videos of a removed class will still be processed, because the call `glob.glob(os.path.join(videos_dir, "*", "*.mp4"))` discovers videos from all available class folders. Can you make sure that this one also only uses registered classes?

corneliusboehm (approved):
:classical_building:

PR #170, opened 2021-05-17
Train global logistic regression model on all available data

corneliusboehm:
Before, one logistic regression would be trained per class and only on the data of the split (test or validation) that was just submitted. Because tags are now globally defined and can be present in multiple classes, it makes more sense to have one global logistic regression model as well. Also we can use both data from training and validation, because this is just used for assisting the user with annotation.

### Details
- `train_logreg` collects data from both splits and all classes
- The call to `compute_frames_and_features` was moved into `train_logreg`, because both calling methods were doing the same operations before
- `train_logreg` was moved into `annotation.py` to prevent the cyclic import `finetuning` -> `utils` -> `finetuning`
- Toggling the `assisted_tagging` flag no longer needs the `split` or `label` information
- The spinner animation was added to all places where there could be a delay because of assisted tagging and its preparation (the assisted tagging checkbox and the annotation navigation menu)
- `compute_frames_and_features` now makes only one call to `get_project_setting`

### DIFF
--- a/sense/finetuning.py
+++ b/sense/finetuning.py
@@ -239,6 +239,8 @@ def compute_frames_and_features(inference_engine: InferenceEngine, project_path:
     :param features_dir:
         Directory where computed features should be stored. One .npy file will be created per video.
     """
+    assisted_tagging = utils.get_project_setting(project_path, 'assisted_tagging')
+
     # Create features and frames folders
     os.makedirs(features_dir, exist_ok=True)
     os.makedirs(frames_dir, exist_ok=True)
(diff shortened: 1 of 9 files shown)


guillaumebrg (commented):
Had a quick look. Everything seems fine to me. 

Tested it on a subset of the SmartTV gesture data and it seems to work.

One user feedback though: When I clicked on "Assisted Labelling", I had annotated ~50 videos already and it started to process those videos on the CPU right ahead, before I had time to check "Use GPU" just below. It took a really long time to process everything on the CPU... should we display "Use GPU" before "Assisted Labelling"? I guess it would be a bit weird... Or would it make sense to activate "Use GPU" by default in case the system has a GPU? I guess there could be another UI trick to fix this. Don't think this should be a blocker for this PR though.

corneliusboehm:
> One user feedback though: When I clicked on "Assisted Labelling", I had annotated ~50 videos already and it started to process those videos on the CPU right ahead, before I had time to check "Use GPU" just below. It took a really long time to process everything on the CPU... should we display "Use GPU" before "Assisted Labelling"? I guess it would be a bit weird... Or would it make sense to activate "Use GPU" by default in case the system has a GPU? I guess there could be another UI trick to fix this. Don't think this should be a blocker for this PR though.

Good feedback. I think the automatic enabling would be the nicest option, so we should add that next.

sunny-panchal (approved):
Everything looks good to me too :+1:

PR #171, opened 2021-05-18
[Feature] Run demos from sensestudio

YasheshSavani:
Description:

This PR will add the functionality of running provided examples in the `sense/examples` directory directly using the user interface in sense studio.

Examples:
1. Calorie estimation
2. Rep counter
3. Fitness tracker
4. Gesture recognition
5. Gesture control


guillaumebrg (commented):
I haven't looked at the code yet but I tested it. It seems to work. Can we increase the size of the video display? It looked like this in my case:

Screenshot from 2021-05-20 10-43-00

The video display should be the biggest element by far in my opinion

YasheshSavani:
> I haven't looked at the code yet but I tested it. It seems to work. Can we increase the size of the video display? It looked like this in my case:
> 
> Screenshot from 2021-05-20 10-43-00
> 
> The video display should be the biggest element by far in my opinion

@guillaumebrg It is because of the original image shape which is coming from the backend after processing is (480, 1140) [height, width], if we want to increase the frame size then we need to do it in `display.py`

sunny-panchal:
The demos work for me as well. When selecting a different demo from the dropdown it keeps playing the previous one. Can we automatically end the inference when the dropdown value changes?

sunny-panchal (commented):
All the demos are working for me now. I'm going to look at the code now but here's some early visual/functional feedback:
- clearing the log when switching to a new task
- Could edit the demo section in the readme
- The dropdowns for `weight` etc. are shown for `run_action_recognition` where it's not needed
- Instead of having two buttons, `run demo` and `cancel`, where `run_demo` is disabled when running, how about a single button that switches between saying `run_demo` and `stop_demo` depending on the state?

tools/sense_studio/static/demos.js
```diff
@@ -0,0 +1,219 @@
...
+document.addEventListener("DOMContentLoaded", function () {
+    toggleMETConverters();
+});
+
+function toggleMETConverters() {
+    let demo = document.getElementById('demo');
+    let metToCalorieConverters = document.getElementById('metToCalorieConverters');
+
+    // Demos containing MET converters (e.g. weight, height, age, gender)
+    // If add new demo with any converter from above, add that script name here in below list (without .py)
```
sunny-panchal, line 11: unclear
```suggestion
    // Some demos need these additional parameters:  weight, height, age, gender
    // NOTE: New demos requiring these should be added to the list below. 
```

tools/sense_studio/static/demos.js
```diff
@@ -0,0 +1,219 @@
...
+    let gender = document.getElementById('gender').value;
+    let modelName = document.getElementById('modelName').value;
+    let webcamInput = document.getElementsByName('inputSource')[0];
+    let saveVideo = document.getElementById('saveVideo');
+    let inputVideoPath = document.getElementById('inputVideoPath');
+    let inputVideoPathValue = (webcamInput.checked) ? '' : inputVideoPath.value;
+    let outputVideoName = document.getElementById('outputVideoName');
+    let outputVideoNameValue = (saveVideo.checked) ? outputVideoName.value : '';
+    let title = document.getElementById('title').value;
+    let buttonRunDemo = document.getElementById('btnRunDemo');
```
sunny-panchal, line 71: Should we stick to the same variable names everywhere? i.e. Changing the element id to also be `buttonRunDemo` to make it easier to track the same reference?

tools/sense_studio/static/demos.js
```diff
@@ -0,0 +1,219 @@
...
+    // Enable run demo button again depending on input fields
+    checkInputFields();
+}
+
+function toggleInputVideoField() {
+    let webcamInput = document.getElementsByName('inputSource')[0];
+    let inputVideoDiv = document.getElementById('inputVideoDiv');
+
+    if (webcamInput.checked) {
+        inputVideoDiv.classList.add('uk-hidden');
```
sunny-panchal, line 144: > Not sure if this is good practice, using the class type to hide the divs that aren't needed. Should be able to wrap it in another function and not build the container altogether.

Same with this.

tools/sense_studio/demos.py
```diff
@@ -0,0 +1,137 @@
+import base64
+import importlib
+import multiprocessing
+import os
+import queue
+
+from typing import Optional
```
sunny-panchal, line 7:
```suggestion
import queue
from typing import Optional
```

tools/sense_studio/demos.py
```diff
@@ -0,0 +1,137 @@
...
+        # Wait until process is complete
+        demo_process.join()
+        demo_process = None
+        stop_event.clear()
+
+    return jsonify(success=True)
+
+
+@socketio.on('stream_demo', namespace='/stream-demo')
+def stream_demo(msg):
```
sunny-panchal, line 99: `msg` unused

sunny-panchal:
I created a small PR #181 with some kwargs refactoring. Other than that, looks good.

sunny-panchal (approved):
Looks good -- thanks for making those changes! 

Note:
- we'll swap run/stop buttons with a single button later
- we'll refactor the `uk-hidden` stuff to be dynamic later

PR #181, opened 2021-06-16
Demos kwargs refactor

sunny-panchal:
Some changes:
- I further updated the kwargs to simplify the function signatures. 
- Typehints throw an error when multiple types are passed in (where `Optional[int, float]`'s were used) -- replaced with just float


YasheshSavani (approved):
LGTM :+1:  Thank you

